PR #706, opened 2012-01-23
added support for server-side encryption on s3

jparker:
Adds an "encryption" attribute to Fog::Storage::AWS::File to enable S3's server-side encryption.


Willardgmoore:
I'm having a hard time trying to figure out implementing sse in my project. I've already got s3 set up working with fog, but every new upload doesn't seem to get that encryption. http://codeshare.io/iIU1c

geemus:
@Willardgmoore I think inside `has_attached_file` you may need to nest the encryption argument inside something to specify it as an arg specifically for fog, but I'm not certain.

PR #1480, opened 2013-01-16
[vsphere] searching for VM improved to search whole cluster instead of current folder.

mjblack:
Sorry for the duplicate, the other one closed when I was trying to rebase an update to the pull request.

The current search options will either only find based on UUID or by VM name. The issue with the VM name is it will only check the top level folder instead of the whole datacenter which means if you use folders for organization the VM will never be found unless at top level. I also moved some functionality to a new function so other functions can take advantage of it without having to duplicate code. Like if I want to just verify that a VM exists and a method can be created to call and verify that a VM was returned as oppose to going through the hash build.

The only caveat is the get_virtual_machine will take a while depending on how many VMs are returned because the attribute of the name appears to not be included as part of the view created and accessing that property looks like it makes another call which slows down the entire process.

raw_list_all_virtual_machines method was created and list_all_virtual_machines calls it so methods like the get_vm_by_name do not have to wait for the hash build from list_all_virtual_machines because the hash build is very slow depending on the VMs returned. An example would be, in my environment we have 600 VMs and just searching for a name takes about ~30 seconds, to build the hash takes about ~4 minutes.

### DIFF
--- a/lib/fog/vsphere/requests/compute/get_virtual_machine.rb
+++ b/lib/fog/vsphere/requests/compute/get_virtual_machine.rb
@@ -3,6 +3,7 @@ module Fog
     class Vsphere
       class Real
         def get_virtual_machine(id, datacenter_name = nil)
+          # The larger the VM list the longer it will take if not searching based on UUID.
           convert_vm_mob_ref_to_attr_hash(get_vm_ref(id, datacenter_name))
         end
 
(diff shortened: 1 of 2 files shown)


nirvdrum:
@mjblack Sorry it took so long for us to get to this.  But it's merged.  I manually resolved the conflicts.  Thanks!

PR #2458, opened 2013-12-06
[vcloud_director] Manage multiple VM connections

astratto:
VMs can have multiple connections, but vm.network retrieved only the first one.

Details about a real response and a correct output are documented in `lib/fog/vcloud_director/parsers/compute/vm_network.rb`

This patch adds options to add, remove and edit VM networks.

### DIFF
--- a/lib/fog/vcloud_director/README.md
+++ b/lib/fog/vcloud_director/README.md
@@ -426,37 +426,83 @@ vm = vapp.vms.get_by_name("DEVWEB")
 vm.network
 ```
 ```ruby
-    <Fog::Compute::VcloudDirector::VmNetwork
+  <Fog::Compute::VcloudDirector::VmNetwork
     id="vm-2ddeea36-ac71-470f-abc5-c6e3c2aca192",
     type="application/vnd.vmware.vcloud.networkConnectionSection+xml",
     href="https://example.com/api/vApp/vm-2ddeea36-ac71-470f-abc5-c6e3c2aca192/networkConnectionSection/",
     info="Specifies the available VM network connections",
     primary_network_connection_index=0,
-    network="DevOps - Dev Network Connection",
-    needs_customization=true,
-    network_connection_index=0,
-    is_connected=true,
-    mac_address="00:50:56:01:00:ea",
-    ip_address_allocation_mode="POOL"
+    connections=[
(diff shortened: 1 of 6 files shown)


nirvdrum:
Sorry I just got around to this.  So, first thing is I've never used vCloud Director.  My experience is solely on the vCenter side.  I'm not sure who the more appropriate maintainer would be.

The one thing that stuck out to me is this change isn't backwards-compatible.  I don't know how many people actually use the vCloud Director stuff inside fog, but as far as I can tell, all of them would be forced to change code when upgrading a minor release of fog.  Generally we try to preserve backwards compatibility and deprecate judiciously, providing that isn't terribly onerous.

nirvdrum:
@geemus Thoughts on my latest comment?

geemus:
@nirvdrum Ah, thanks for the clarification. A breaking change in a minor is definitely not desirable. Unfortunately I also do not know whom best to call upon to check this out.

astratto:
I'm sorry guys it took me so long to reply...

@nirvdrum is right that this change is not backward-compatible, but at the same time the current behaviour (or at least the behaviour we had one year ago...) is wrong because it assumes that a VM can only have one network assigned.
This constraint is not acceptable in many scenarios, hence this PR.

Another more conservative approach could be:
- introduce the new behaviour while wrapping and deprecating the old one in the next minor
- get rid of the old stuff in another minor

Sadly, I don't have access anymore to a real vCloud Director instance, so I won't be able to rebase and adapt this PR anytime soon...

nirvdrum:
@mikepea It looks like you might be the go-to guy for vcloud_director.  If that's accurate, do you mind taking over from here?  I'm going to unassign myself.

plribeiro3000:
@astratto could you resolve the merge conflicts?

astratto:
@plribeiro3000 sorry, but as I mentioned I don't have a way to verify that code against a real vCloud Director instance

plribeiro3000:
@astratto Ok. So are you saying there is no way to keep up with the work done here?

astratto:
@plribeiro3000 "there is no way" is a little bit extreme... :D 

You can safely assume that at the moment I'm not able to work on that though.

plribeiro3000:
@astratto hahah. You are right. Sorry for those words. :sweat: 

Alright, maybe someone else can assume it?

astratto:
@plribeiro3000 no problem, I just wanted to set the correct expectations... ;)

plribeiro3000:
Closing by lack of progress. Feel free to reopen or comment back on it to continue discussion.

geemus:
Thanks!

PR #2998, opened 2014-06-16
[hp] reauthenticate

gogolok:
The current hp (storage) provider doesn't support token expiration.

We have a token expiration time of around ~50 minutes when using our Swift service.
Afterwards the fog gem will throw an exception.

http://docs.openstack.org/api/quick-start/content/
```
2. Send API requests and include the token in the X-Auth-Token header. Continue to send API requests with that token until the job completes or a 401 Unauthorized error occurs.
3. If the 401 Unauthorized error occurs, request another token.
```

This PR fixes the hp (storage) provider.

When creating the request it's important the original `params` won't be modified so that the `retry` will work.

### DIFF
--- a/lib/fog/hp/storage.rb
+++ b/lib/fog/hp/storage.rb
@@ -303,55 +303,37 @@ module Fog
         attr_reader :hp_cdn_ssl
 
         def initialize(options={})
+          @options = options
+
           # deprecate hp_account_id
-          if options[:hp_account_id]
+          if @options[:hp_account_id]
             Fog::Logger.deprecation(":hp_account_id is deprecated, please use :hp_access_key instead.")
-            options[:hp_access_key] = options.delete(:hp_account_id)
+            @options[:hp_access_key] = @options.delete(:hp_account_id)
           end
-          @hp_access_key = options[:hp_access_key]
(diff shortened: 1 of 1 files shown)


geemus:
@mwhagedorn thanks, good to know.

lib/fog/hp/storage.rb
```diff
@@ -303,55 +303,37 @@ class Real
...
-          else
-            # Call the legacy v1.0/v1.1 authentication
-            credentials = Fog::HP.authenticate_v1(options, @connection_options)
-            # the user sends in the cdn endpoint
-            @hp_storage_uri = options[:hp_auth_uri]
-            @hp_cdn_uri  = options[:hp_cdn_uri]
-          end
-
-          @auth_token = credentials[:auth_token]
+          authenticate(@options)
```
plribeiro3000: Why do you need to pass `@options` here?
I think you can just use `@options` in the context of `authenticate`, right?

plribeiro3000:
@gogolok Could you rebase this PR again please?

plribeiro3000:
Closing due inactivity. Comment here when you have time to work on top of this again.

geemus:
Thanks!

PR #3061, opened 2014-07-17
[openstack] evacuate server implementation

sabcio:
Evacuate server to make them available again on different host. 

Supports onSharedStorage, adminPassword options.

### DIFF
--- a/lib/fog/openstack/compute.rb
+++ b/lib/fog/openstack/compute.rb
@@ -82,6 +82,7 @@ module Fog
       request :get_vnc_console
       request :live_migrate_server
       request :migrate_server
+      request :evacuate_server
 
       # Image CRUD
       request :list_images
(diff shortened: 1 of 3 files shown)


plribeiro3000:
@sabcio Could you try to add tests around this changes?

Ladas:
oh man, I haven't noticed this, I did the same PR https://github.com/fog/fog/pull/3799

tests for this will be kind of tricky, since it requires at least 2 Compute hosts setup, which the development setups are not, right? Best would be to move this into VCR recording and just checking the correct API is called.

Temikus:
I think this has been added in already (Presumably by #3799):
https://github.com/fog/fog-openstack/blob/71bd4ec60980e0413126ff24c2559f9992386714/lib/fog/openstack/requests/compute/evacuate_server.rb

@plribeiro3000 - can you verify? If so - please, close.

Ladas:
@Temikus @plribeiro3000 yeah, we use evacuate server, so it should be working in fog-openstack

plribeiro3000:
So we can close this then.

Thanks for your feedback!

PR #3098, opened 2014-08-10
Fixed incorrect docs for using tags for autoscaling groups

endzyme:
Tags previously recommended using an array but each_with_index didn't
work properly. Previously it would jumble all your hash into a value if
you used this inline documentation.


plribeiro3000:
@endzyme Can you create this patch against [fog-aws](https://github.com/fog/fog-aws)?

Thanks!

geemus:
Thanks!

endzyme:
Leaving it hanging due to lack of free time :[ sorry!

PR #3117, opened 2014-08-19
[vsphere] get resource pool without name

teaforthecat:
This appears to be the best way to allow cloning into a chosen cluster which is not using any resource pools. Luckily the cluster is able to return a default resource pool. I have this working with vsphere 5.1 and rbvmomi 1.6.0.
The option to the vm_clone method would look like this `'resource_pool' => ['cluster5', nil]`

### DIFF
--- a/lib/fog/vsphere/requests/compute/get_resource_pool.rb
+++ b/lib/fog/vsphere/requests/compute/get_resource_pool.rb
@@ -13,7 +13,7 @@ module Fog
         def get_raw_resource_pool(name, cluster_name, datacenter_name)
           dc      = find_raw_datacenter(datacenter_name)
           cluster = dc.find_compute_resource(cluster_name)
-          cluster.resourcePool.find name
+          name.nil? ? cluster.resourcePool : cluster.resourcePool.find( name)
         end
       end
 


teaforthecat:
Excellent analysis. To be clear, I'm trying to clone to a cluster that the template is not on, as well as use the default resource pool (or there are no named resource pools). So the template is on the "prod" cluster for example, and I want to clone it to "non-prod". I believe the issue is that  `cluster.resourcePool.find()` does not except nil as an argument, but I need to verify that, it's been a while. I'll get back shortly. Thanks.

nirvdrum:
Thanks.  As you can see, we don't have the best test suite for vSphere stuff, so I try to 100% understand the problem and proposed solution before merging PRs.  It takes longer, for sure, but I really don't want to introduce any potential regressions.

nirvdrum:
Great.  I'll try to recreate the situation in my own vSphere installation to better understand the workflow.

teaforthecat:
Hi @nirvdrum , I was able to replicate the issue to confirm the motivation for this pull request. When cloning a template from one cluster to another without a resource pool name I get this error:
```
[2015-02-04 09:07:33 -0600] [pid: 30048] [10.10.3.13] missing required parameter name
  [..........]
  /opt/rh/ruby193/root/usr/share/gems/gems/rbvmomi-1.6.0/lib/rbvmomi/vim/ResourcePool.rb:6:in `find'
  /opt/rh/ruby193/root/usr/share/gems/gems/fog-1.21.100/lib/fog/vsphere/requests/compute/get_resource_pool.rb:19:in `get_raw_resource_pool'
```
So the check for a nil name is required.

nirvdrum:
Thanks for the clarification!  Merged.

geemus:
Thanks!

PR #3121, opened 2014-08-20
[vsphere] find network by name and dvswitch

teaforthecat:
given a name and a switch name
or just a name and whether a dvs is desired
or just a name

When you just want the first (or only) distributed virtual switch on a given cluster, you for example call `get_raw_network(name, datacenter_name, true)` This enables a network admin to choose where a new machine will be placed on the network by moving a network to a distributed virtual switch or non distributed virtual switch.

### DIFF
--- a/lib/fog/vsphere/requests/compute/get_network.rb
+++ b/lib/fog/vsphere/requests/compute/get_network.rb
@@ -10,15 +10,44 @@ module Fog
 
         protected
 
-        def get_raw_network(name, datacenter_name, distributedswitch_name=nil)
-          dc = find_raw_datacenter(datacenter_name)
-
-          @connection.serviceContent.viewManager.CreateContainerView({
-            :container  => dc.networkFolder,
-            :type       =>  ["Network"],
-            :recursive  => true
-          }).view.select { |n| n.name == name and (not distributedswitch_name or n.config.distributedVirtualSwitch.name == distributedswitch_name)}.first
+        def get_raw_network(name, datacenter_name, distributedswitch=nil)
+          finder = choose_finder(name, distributedswitch)
+          networks = get_all_raw_networks(datacenter_name)
+          networks.find { |n| finder.call(n) }
         end
+      end
+
+      module Shared
+
+        protected
+
+        def get_all_raw_networks(datacenter_name)
+          dc = find_raw_datacenter(datacenter_name)
+          @connection.serviceContent.viewManager.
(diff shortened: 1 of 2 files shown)


teaforthecat:
Hi @nirvdrum , thanks for reviewing this. I tried to address most of the concerns you raised. I came up short on a solution to the `choose_finder` method using polymorphism, but hopefully the `:dvs` option is more clear than the `true` option.  Since it's been awhile (and I was a little push happy), should I rebase this branch? I think there is an unrelated failing test.

nirvdrum:
Ahh, sorry, I had missed the updates.  I'm traveling for work this week.  I'll try to give it another look, but I'm planning on doing a fog spike when I get back and this will probably get pulled into there.

nirvdrum:
@teaforthecat Okay, the changes here look much nicer to me.  I'm not usually one to call for rebases, but consolidating these commits would be nice.

A few general comments:

* I've lost my access to vSphere licensed for distributed switches.  I can create them and port groups in the version I have, but I can't assign any hosts to them and thus any VMs.  So I can test a good chunk of the API interaction, but not all of it through-and-through.
* I'm pretty comfortable merging this now, pending the minor clean-up.
* While I could create a non-distributed switch and a distributed switch and a port group all with the same name, I couldn't create two port groups with the same name, even across switches.  I'm not sure if I'm just testing wrong, but it seems to suggest the cases here are a bit simpler than I thought: exact DVS match, the only DVS with the network name (rather than the 'first' one), and a non-DVS network.
* It looks like the path this goes down helps the create_vm task and that one alone.  That's fine, but it'd be nice if the model layer supported similar network searching.  As it stands, that uses :get_network, which ignores the DVS option altogether.  It's a bit out of scope here anyway, but it'd be nice to have for future work.

teaforthecat:
I did the deed; rebased to clean up the branch.

nirvdrum:
Thanks!  I just merged the change. But if you get a spare moment, could you please address whether my synopsis about the logic cases is accurate (the 3rd bullet point in https://github.com/fog/fog/pull/3121#issuecomment-72564993)

teaforthecat:
To comment on the 3rd bullet point above - I haven't really worked with port groups, but that sounds right to me; no two port groups with the same name across switches. And we agree upon the three-branch scenario you mention. - Thanks for spending time on this.

geemus:
Thanks!

PR #3122, opened 2014-08-20
[openstack] support filtering in list servers mock

danleyden:
Make the mocking for openstack closer to reality


plribeiro3000:
@danleyden Could you open this against `fog-openstack` please?

Thanks!

PR #3123, opened 2014-08-20
[openstack] deprecate namespaced params

blentz:
This normalizes all of the class options for the openstack provider by removing the prepended "openstack_" string to all config parameters.

When developing tooling for a multi-cloud environment (e.g. openstack and aws), the goal should be to abstract away differences between cloud providers. This is not possible when each fog provider has chosen to add provider-specific nomenclature to their configuration parameters.

When I specify "region", it should be implied that it means "region" in whatever provider I've instantiated. I shouldn't have to add complexity to my code to say "In the AWS case, the param is called "region", but in the OpenStack case, it's called "openstack_region".

This pull request is aimed at fixing this issue.

### DIFF
--- a/lib/fog/openstack/compute.rb
+++ b/lib/fog/openstack/compute.rb
@@ -3,13 +3,21 @@ require 'fog/openstack/core'
 module Fog
   module Compute
     class OpenStack < Fog::Service
+      # the openstack_* params are all deprecated. remove the
+      # 'openstack_' part of the string from your config parameter names.
       requires :openstack_auth_url
+      recognizes :auth_url
+
       recognizes :openstack_auth_token, :openstack_management_url,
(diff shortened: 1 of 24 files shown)


plribeiro3000:
@blentz Take a look at #3348

Temikus:
@blentz - I think the discussion should be moved to https://github.com/fog/fog-openstack
Since `fog-openstack` has drifted quite a bit since your PR, I'm afraid this will need to be reopened in the metagem.

/CC @plribeiro3000 - I think this needs to be closed - WDYT?

plribeiro3000:
:+1:

Em ter, 28 de jun de 2016 06:53, Artem Yakimenko <notifications@github.com>
escreveu:

plribeiro3000:
Closing. Thx!

Issue #3134, opened 2014-08-31
[vsphere] vm_clone method uses wrong path to find vm


teaforthecat:
If you pass 'template_path' but not a 'dest_folder' in the options to the vm_clone method, the vm will be created in the template folder but the get_virtual_machine method at the end of the vm_clone method uses the option['dest_folder'] which is nil so the path is wrong.

PR #3139, opened 2014-09-02
[vsphere] new default dest_folder in vm_clone

teaforthecat:
fixes fog/fog#3134

We want to allow someone to not set the dest_folder option because they
might not have a desired folder structure and not care. In case they
follow common examples and use a folder called 'templates/' for templates
we don't want to create the new vm in that folder. It is better to
default to the root folder.

Conflicts:
	lib/fog/vsphere/requests/compute/vm_clone.rb

### DIFF
--- a/lib/fog/vsphere/requests/compute/vm_clone.rb
+++ b/lib/fog/vsphere/requests/compute/vm_clone.rb
@@ -82,8 +82,8 @@ module Fog
 
           # Options['dest_folder']<~String>
           # Grab the destination folder object if it exists else use cloned mach
-          dest_folder = get_raw_vmfolder(options['dest_folder'], options['datacenter']) if options.key?('dest_folder')
-          dest_folder ||= vm_mob_ref.parent
+          dest_folder_path = options.fetch('dest_folder','/') # default to root path ({dc_name}/vm/)
+          dest_folder = get_raw_vmfolder(dest_folder_path, options['datacenter'])
 
           # Options['resource_pool']<~Array>
           # Now find _a_ resource pool to use for the clone if one is not specified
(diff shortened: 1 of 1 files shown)


lib/fog/vsphere/requests/compute/vm_clone.rb
```diff
@@ -82,8 +82,8 @@ def vm_clone(options = {})
 
           # Options['dest_folder']<~String>
           # Grab the destination folder object if it exists else use cloned mach
-          dest_folder = get_raw_vmfolder(options['dest_folder'], options['datacenter']) if options.key?('dest_folder')
-          dest_folder ||= vm_mob_ref.parent
+          dest_folder_path = options.fetch('dest_folder','/') # default to root path ({dc_name}/vm/)
```
teaforthecat: It is actually that I put my templates in a "templates" folder and I don't want my vm's to be cloned into that folder.
In the line below, I found that the `convert_vm_mod_ref_to_attr_hash` was needed because `get_virtual_machine` would fail to find the vm.

nirvdrum:
@teaforthecat Thanks for the follow-up.  I've finally just gotten around to trying this out.  The changes you made aren't working for me, so I need to dig through that a bit.  But, is there a reason you can't just set dest_folder when cloning?  Setting it to '/' should give you the behavior you're looking for.  I'm a bit worried about changing the default behavior in the case of not specifying anything.  I think I agree '/' is a more sensible default . . . I'm just not sure if anyone is relying on the old behavior.

nirvdrum:
Nevermind, I screwed up.  The patch works fine.

I'm still a bit reticent to change a default, but I think the change makes sense.  I'll roll with it and see if anyone files a bug in changing behavior.  I'm partially inclined to do this because I couldn't get fog to work properly when the 'dest_folder' option wasn't set, but also didn't default to '/'.

nirvdrum:
Thanks again!  And sorry it took so long to get to this.

nirvdrum:
Well, I've already merged it.  But for posterity, the issue I described with not finding the VM seems to be address in #1480.

geemus:
Thanks!

PR #3186, opened 2014-09-29
adding support for network interfaces and public IP association (non-default VPCs) for EC2 spot instances

CpuID:
As per subject. There were no tests around this stuff right now, and no mocks. I don't feel super confident implementing them myself without some coaching though... let me know if this is an issue.


plribeiro3000:
@CpuID I'm closing this since aws code now resides at [fog/fog-aws](https://github.com/fog/fog-aws). Could you make a patch to that repo instead? Thanks!

geemus:
Thanks!

PR #3233, opened 2014-11-02
Ec2 eu central

fcheung:
This adds support to for the eu-central-1 region (ec2 only). This requires the following changes:
- adding eu-central-1 to the whitelist of regions (I noticed some services had their own whitelist - I assume this is an artifact of times when certain services only existed in some regions)
- switching over to user aws signature v4

All AWS regions support AWS signature v4 (new regions, such as eu-central-1 only support v4), so this won't break anything for existing AWS users. I'm less sure about the impact on people using services with an AWS compatible API (eg eucalyptus) - while I assume current versions support v4, presumably some people out there still use older versions.

If we can agree the approach (eg should we allow people to request signature v2?) then it should be straightforward to make this change to the other aws services

### DIFF
--- a/lib/fog/aws/requests/storage/copy_object.rb
+++ b/lib/fog/aws/requests/storage/copy_object.rb
@@ -28,7 +28,7 @@ module Fog
         # @see http://docs.amazonwebservices.com/AmazonS3/latest/API/RESTObjectCOPY.html
         #
         def copy_object(source_bucket_name, source_object_name, target_bucket_name, target_object_name, options = {})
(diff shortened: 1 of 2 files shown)


nottinhill:
We need Fog x.24 for Sluice x.2.2. Is it possible to get the EU-Central Feature into x.24? If not, can anybody point us to the correct file that needs to be changed, if it is even that easy (from above thread I suppose it isn't)

fcheung:
You do basically need the entirety of that commit, although if you only need some services then you could just update the files for that service, in addition to the stuff in core. You might also need some of the subsequent changes (that have occurred in fog-aws), there have been some changes to the s3 signing.

PR #3303, opened 2014-11-29
[AWS|Storage] allow signature version to be set to 2 (WIP)

fcheung:
I've had a stab at this, just for main requests (doesn't support presigned urls yet). There's some fiddly stuff going on - signature v4 requires that you do canonicalisation of path/host etc before you sign, signature v2 behaves differently, signature v4 requires that you regenerate the signature when following a redirect but signature v2 doesn't, so there are quite a few possible combinations to test.

I think the other amazon services would be a lot easier to add support for multiple signature versions to.

### DIFF
--- a/lib/fog/aws/storage.rb
+++ b/lib/fog/aws/storage.rb
@@ -488,10 +488,11 @@ module Fog
           params[:headers] = (params[:headers] || {}).dup
 
           params[:headers]['x-amz-security-token'] = @aws_session_token if @aws_session_token
-          params[:headers]['x-amz-date'] = date.to_iso8601_basic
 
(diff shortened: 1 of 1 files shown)


ramonskie:
any idea when the next fog release is where this is added?

fcheung:
I think this was in fog 1.26 (and you can now use just the fog-aws gem, which definitely includes all of this)

ramonskie:
i'm looking at some commits but its one big mess
i see there is allot of stuff moved around.
so i hope someone can explain to me which version of which gem i need :)

fcheung:
This was merged before fog was split off into fog-aws

PR #3309, opened 2014-12-03
Add ability to create catalogs in vcloud

ConnorAtherton:
This is my first time contributing so if there is anything I've done incorrectly please let me know. It works but I would like to add some test cases. It's late so I will review how fog already implements things like mocks tomorrow and see what I can do.


geemus:
Hey, thanks for closing this out (as imagine it's a bit out of date). Just let me know if you have questions or anything further we can help with though.

PR #3310, opened 2014-12-03
[DigitalOcean] Image full name & arch helpers

dLobatog:


seanhandley:
Given that API v1 is deprecated within weeks, and that this has been unmerged for 10 months...

geemus:
Good call, thanks.

PR #3328, opened 2014-12-10
Include ovirt_ca_no_verify option.

pparkkin:
Add support for rbovirt ca_no_verify connection option through ovirt_ca_no_verify.

I tried to look around at the test cases to figure out how to add a test case for the new option, but couldn't find a good way to do that. No tests added. All current mock tests for ovirt pass.


geemus:
No problem, good to get this fixed.

Issue #3099, opened 2014-08-11
Update Nokogiri version from 1.5.11 to 1.6


tnimmo:
Are there any plans to upgrade Nokogiri to version 1.6?

I have forked fog and upgraded nokogiri to 1.6 and all seems to be working. Should I put this in as PR or are there other considerations I haven't thought about?

From the Nokogiri change logs (1.6.0.rc1):

"Ruby 1.9.2 and higher now required"
"Support for Ruby 1.8.7 and prior has been dropped"

Is this a problem?

isabanin:
With fog being dependent on Nokogiri 1.5 it's currently not possible to use it in a Rails 4.2 application, which requires Nokogiri 1.6.

geemus:
@isabanin thanks for the note. Maybe our hand is going to be forced here. I need to get on top of 2.0 plans anyway...

tokengeek:
I think this ticket belongs in https://github.com/fog/fog-xml

I actually thought this was no longer the case and that 1.8.7 users were specifying it in their application `Gemfile`s as they have to for a number of other dependencies.

`fog` will need an update to the 1.8.7 Gemfile to avoid breaking on travis however.

geemus:
Ah, that sounds a good deal more promising.

vlasar:
Any update on being able to use ```fog``` with Rails 4.2?

Do we need to wait for 2.0 version or did I misunderstood?

Thank for all your work.

plribeiro3000:
@viasar Since `ruby 1.8.7` will be dropped on `fog 2.0.0` we can't bump the `nokogiri` dependency until it happens.

@geemus Perhaps we should hurry a bit on `fog 2.0.0` to add compatibility with `rails 4.2.0`?

vlasar:
@plribeiro3000 Ah, makes sense. Thank you for the clarification.

tokengeek:
I don't think there's much holding us up (except for available time).

We can bump the version numbers and merge the PR. It's just making sure
all the critical things (if any) are done.

agis-:
Any news on this one? It's really blocking us from going to Rails 4.2 too.

plribeiro3000:
I think it past time to move on with this. Let me sync to see what we need to accomplish to be sure not to break anything else.

I will let you know once we do it.

Thanks for your support!

agis-:
@plribeiro3000 I'd be happy to help with this in order to speed things up. Let me know :)

tokengeek:
[...]

PR #3348, opened 2014-12-17
[Fog 2.0] Remove 1.8.7 support

tokengeek:
This is the official commit to drop Ruby 1.8.7 support from fog.

This is a backwards incompatible change (for hopefully obvious reasons)
so will be part of the next MAJOR fog release `v2.0`

Ruby 1.8.7 is no longer supported by the language developers.

Anyone needing to stay on Ruby 1.8.7 such as users of LTS distros should
remain on the `v1.x` series of fog.


calebthompson:
1.9.3 is also EOL. https://www.ruby-lang.org/en/news/2015/02/23/support-for-ruby-1-9-3-has-ended/

geemus:
Yeah. I think we should drop it, but we will probably wait for a major version bump to do it.

calebthompson:
Sorry, maybe I misunderstand. Isn't this a branch for a major version bump?

geemus:
Oh, right, sorry. My brain was still thinking about the travis related matrix changes...

calebthompson:
Right. And _that_ PR was to this branch :D

geemus:
AH. I totally missed that context I guess.

tokengeek:
Okay, apart from the conflict that'll appear when bumping the version in `fog.gemspec` this is probably ready for merging.

So when merged and 1.8.7 support is officially dropped we can bump dependencies (#3099) and people can test with later Rails versions.

agis-:
So any estimates on this? Thanks.

tokengeek:
@agis- @geemus should be doing a release in the next few hours for the last 1.x release. Merging this then releasing a pre release version of 2.0.0-alpha for testing.

So by the end of Wes' day I hope.

geemus:
Thanks!

PR #3364, opened 2014-12-23
Add support for bucket notification configuration.

cyberdelia:
Add support for the new s3 notifications configuration, see:
http://docs.aws.amazon.com/AmazonS3/latest/API/RESTBucketGETnotification.html
http://docs.aws.amazon.com/AmazonS3/latest/API/RESTBucketPUTnotification.html


geemus:
Looks like a great start. Could you add tests around the new functionality? Also, could you add mock implementations (fine if they just raise not implemented) for completeness? Thanks!

geemus:
Awesome, thanks!

PR #3376, opened 2015-01-02
fog-aws module

lanej:
/cc @geemus


plribeiro3000:
Great job @lanej! :clap: 

Can you also add @geemus as contributor on the gem?

Also, does AWS make calls using xml and json?

plribeiro3000:
Just in case you need look at this link: http://guides.rubygems.org/command-reference/#gem-owner

lanej:
@plribeiro3000 i found several references to `Fog::JSON` https://github.com/fog/fog-aws/blob/27c1bf05c7ed6775fd651c73427050a702abbccf/lib/fog/aws/parsers/iam/get_group_policy.rb#L18

Added geemus as a gem owner and into the gemspec

plribeiro3000:
@lanej Interesting. Can you take a look at the broken build?

lanej:
@plribeiro3000 i screwed up the aws bin. should be fixed by https://github.com/lanej/fog/commit/1e4f60d4a734fc32dfd05411cee675c1bf4c6473

plribeiro3000:
@lanej The bin should be inside `fog-aws`.

Once you get it there, the tests here should pass.

lanej:
@plribeiro3000 i was referring to https://github.com/fog/fog/blob/master/lib/fog/bin.rb https://github.com/fog/fog/blob/master/lib/fog/bin/aws.rb.

plribeiro3000:
I'm talking about your commit lanej/fog#1e4f60d4a734fc32dfd05411cee675c1bf4c6473. This file should be inserted inside `fog-aws`.

lanej:
@plribeiro3000 the `Fog::Bin` module is defined in `fog/fog`.  Do you mean I should move `lib/fog/bin/aws.rb` into `fog-aws` and require it in `lib/fog/bin.rb` ?

plribeiro3000:
@lanej Yep.

lanej:
@plribeiro3000 will do.

plribeiro3000:
You should let the original code as it is. All the references must be the same to keep compatibility. The only difference is that the files will be outside the `fog` gem.

lanej:
@plribeiro3000 i think https://github.com/lanej/fog/commit/2558acbd5755d39d0f64c11988717175b394fccd and https://github.com/fog/fog-aws/commit/9a3349b6029b09e52e0e65b3b16d3ee0bcbe756d should take care of it.

plribeiro3000:
Why is your `aws` bin differently from thew one in master of `fog` ?

https://github.com/fog/fog/blob/master/lib/fog/bin/aws.rb

lanej:
@plribeiro3000 because the superclass of `Fog::Bin` is not available in `fog-core` but the `AWS[]` syntax is used in tests.

plribeiro3000:
@lanej This is a know problem and we are looking towards to fix it. @tokengeek has more insight related to this.

The short solution is: The bin just works when you require `fog` and because of that everything is loaded. Bin will not work until we fix this.

lanej:
@plribeiro3000 understood.  is this workaround acceptable?

also, it appears that the jruby builds are failing but i don't believe it is related to this branch.

plribeiro3000:
If the behavior and features will remain the same i think it is ok. Will the bin still work as expected?

lanej:
@plribeiro3000 i will check but the interface is the same so it should.

plribeiro3000:
@lanej I will appreciate if you can check it.

After this i believe that my job in here is done. Maybe @fcheung could have something more to say?

lanej:
looks good

```
jlane@lanej p/fog [fog-aws] » ./bin/fog aws                                                            ‹2.1.5p273›[1]
[fog][WARNING] Unrecognized service: orchestration
[fog][WARNING] Unrecognized service: networkingV2
[fog][WARNING] Unsupported StormOnDemand service: compute
[fog][WARNING] Unsupported StormOnDemand service: network
[fog][WARNING] Unsupported StormOnDemand service: storage
[fog][WARNING] Unsupported StormOnDemand service: dns
[fog][WARNING] Unsupported StormOnDemand service: billing
[fog][WARNING] Unsupported StormOnDemand service: monitoring
[fog][WARNING] Unsupported StormOnDemand service: support
[fog][WARNING] Unsupported StormOnDemand service: models
[fog][WARNING] Unsupported StormOnDemand service: vpn
  Welcome to fog interactive!
  :aws provides AWS, DNSimple, OpenNebula, Openvz and Vmfusion
>> Fog::AWS
Fog::AWS
>> Fog::Compute::AWS.new
#<Fog::Compute::AWS::Real:70231743780220
```

plribeiro3000:
@lanej :+1: 

I just restarted the jruby builds but i guess we might need to create a custom Gemfile to use with it.

lanej:
@plribeiro3000 jruby still not working. not sure what the deal is. i don't see what depends on `curses`.

plribeiro3000:
`curses` has C bindings. It won't work with Jruby. Since you did not add this dependency i think we can skip it for now.

Since the last build in [master](https://travis-ci.org/fog/fog/builds/45495694) was successfull i guess the problem relates to a fog dependency breaking. It is something from the last 2 days. =s

fcheung:
Seems to behave in some (admittedly casual) tests

tokengeek:
Re: the "bin" stuff.

I'm deprecating that behaviour soon, using a toplevel `Provider` namespace eg. `AWS[:service]`

The PR for deprecation has been sitting around for a while and I expect to remove it from fog 2.0.

Basically doing `require "fog"` but having `AWS` as a top class construct is messy and namespace polluting. It basically breaks the entire abstraction layer/point of fog.

The problem (which I why I didn't push the PR) is that as mentioned in the comments above the AWS tests were written the wrong way and deprecating (with a warning) floods the test output.

So the AWS tests need fixing (first!) and that bin file should return to `fog` before this is merged. Basically 2558acbd5 can probably be reverted.

Anyone requiring just `fog-aws` SHOULD NOT use the `AWS` namespace just `Fog::AWS` so they need to use `fog` and fix deprecations beforehand but the support tooling isn't there yet.

Since AWS is the big one with lots of potential for complaints, it needs the most careful refactoring.

I'll try to get that PR done pushed today but it's `core` updates and releases first.

tokengeek:
See https://github.com/fog/fog/issues/3378 for notes about the Travis failures. Seems to be a Travis platform issue.

lanej:
I will

* revert the bin changes
* fix the tests that reference the top-level AWS constant

lanej:
rebased on top of https://github.com/fog/fog/tree/eliminate_toplevel_bin

lanej:
Ready to go when #3377 is merged

tokengeek:
@lanej You might need to bump the dependencies so `fog` is
looking for `fog-aws` with an extracted service mapper.

I assume it will work in the mean time but it might be better to just
do it now in `fog-aws`

lanej:
@tokengeek i yanked all the previous versions that were broken.

lanej:
@fog/owners i'd like to get this merged as soon as possible.  every time some body makes an aws change in fog/fog I'm going to get a conflict.

#3377 should probably get merged first.

geemus:
Diff is huge and my browser doesn't want to play nice. The numbers look right though and some others have reviewed, so bringing this in. We can try to move issues/pulls across as we can going forward. Thanks!

lanej:
woohoo ! thanks @geemus

geemus:
Thank you!

plribeiro3000:
Thank you @lanej. This step you just made are really gonna help us toward our next goal! :+1:

PR #3377, opened 2015-01-03
[AWS] Move key to class mapping to separate file

tokengeek:
This was triggered by #3376

This is a workaround for moving the provider specific code into `fog-aws`
whilst not cluttering the module with the "bin" behaviour.

The bin behaviour usually relates to starting `fog` from the command line
and using it interactively when having a `AWS` class seemed like a nice idea.

The `AWS` namespace should rightly be available to the AWS SDK (uses `Aws`)
so this is being deprecated however each implementation has been done
standalone.

Each top level class basically includes lookup tables between a key and the class
that should be returned.

Whilst I am working on fixing the provider/service registration in `fog-core` this
is a potential fix to put these back into the modules control.

I foresee that `lib/fog/bin/aws.rb` will disappear in 2.0 and `Fog::AWS::ServiceMapper`
will have it's functionality replaced by core components so also likely to go.

/cc @lanej @plribeiro3000
/cc @mdarby - I think this may also be a short term solution for `fog-rackspace`

### DIFF
--- a/lib/fog/aws.rb
+++ b/lib/fog/aws.rb
@@ -1,3 +1,4 @@
+require 'fog/aws/service_mapper'
 require 'fog/aws/auto_scaling'
 require 'fog/aws/beanstalk'
 require 'fog/aws/cdn'
--- a/lib/fog/aws/service_mapper.rb
+++ b/lib/fog/aws/service_mapper.rb
@@ -1,0 +1,129 @@
+module Fog
+  module AWS
+    # @api private
+    #
+    # This is a temporary lookup helper for extracting into external module.
+    #
+    # Cleaner provider/service registration will replace this code.
+    #
+    class ServiceMapper
+      def self.class_for(key)
+        case key
+        when :auto_scaling
+          Fog::AWS::AutoScaling
+        when :beanstalk
+          Fog::AWS::ElasticBeanstalk
+        when :cdn
+          Fog::CDN::AWS
+        when :cloud_formation
+          Fog::AWS::CloudFormation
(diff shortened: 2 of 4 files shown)


tokengeek:
Force pushed rebase to pick up ignoring the broken JRuby builds

geemus:
Sounds like a plan. Would be curious to know how much bin stuff even gets used. Maybe it is more trouble than it is worth and we should consider broader deprecation? Hard to know beyond anecdotal evidence though.

tokengeek:
Pretty much it's https://github.com/fog/fog/blob/master/lib/fog/bin.rb

So it's used by the `Fog` class to expose a few things at the
higher level which is used by the Shindo tests.

It's one of the main reasons why I didn't want these bin files in core
since it is only used by `fog` on the command line except where
others have begun depending on it.

So +1 for deprecating it. This is the work around. Ideally provider
registration would handle all of this but short term if extracted
code managed this lookup for now, it means they can keep this
list up to date from their gems.

The `Fog.credentials` global is still there so this stuff needs
rewriting before it is generic enough anyway.

Basically if a service is `available?` (as a Class) depends on the
global state of the `.fog` file with this code otherwise I'd have
pushed most of `available?` down to the provider code.

Whilst Shindo is kind of deprecated, it's still the majority of testing
so we need to get the helper cleaned up to test providers without
hitting code like this I intend to rework.

But for now I'm happy with this as a template to help.

geemus:
Yeah, seems a good stopgap at least.

geemus:
Thanks!

PR #3384, opened 2015-01-08
Add OpenStack Ironic support

Ladas:
Adding Openstack Ironic support for nodes, chassis, ports and
drivers.

Adding only basic CRUD operations support for start. Rest of
the special methods will be added later. All not implemented
API calls had been marked as TODO.

DependsOn: https://github.com/fog/fog-core/pull/117


geemus:
Real tests should be like `FOG_MOCK=false bundle exec shindo tests/openstack` I believe.

Ladas:
Still the same result with 
FOG_MOCK=false bundle exec shindo tests/openstack
as above, not sure what I am missing.

geemus:
You may need to ensure that the test mock helper has openstack credentials as you expect here: https://github.com/fog/fog/blob/master/tests/helpers/mock_helper.rb (it skips tests for services it doesn't believe it has creds for).

Ladas:
Ah, but this https://github.com/fog/fog/blob/master/tests/helpers/mock_helper.rb is the mocked branch

So if I put credentials there I need to do
export FOG_MOCK=true
bundle exec shindo tests/openstack

Then it does the tests and succeeds, though it is not running with the real APIs, since I can put there any credentials.

So with FOG_MOCK=false, I would expect it should take credentials from ~/.fog ?

Ladas:
Still no luck running shindo tests against real API. But I tested it manually and it works, also tested it in ManageIQ and we are collecting data.

geemus:
Sorry, yes. I think my brain did some of that backwards. The test helper there should provide for running mocked tests, otherwise I believe it should pull from ~/.fog credentials (and skip anything it can't find). I think it will use credentials nested under the `default` key inside that file. But you can specify a different key and/or different file with ENV vars, see: https://github.com/fog/fog/blob/master/tests/helper.rb#L1

Hope that helps, hopefully we can get that running and go from there.

geemus:
P.S. I gather the failing tests on here are the ones that require that fog-core merge?

Ladas:
Yes the failing tests needs the fog-core dependency.

I think I will need to investigate the shindo tests, it doesn't do what it should. I should have the correct credentials, since all works when I use fog console. Also when I run all shindo tests it doesn't skip the ones without credentials but fails on not having credentials.

Ladas:
Are the non working shindo tests with real API blocker for this patch?

Cause other than that, it should be ready with working mocked tests and it works for me nicely when used in ManageIQ.

geemus:
I would feel more confident if the real tests had been run, though the other parts are a good step in the right direction.

@tokengeek could this related to the registered_providers stuff?

Ladas:
I guess the shindo tests can discover, that some attribute has wrong type, otherwise it should all work.

So if we don't need to block this patch for that reason I would be glad. :-) So far I have no idea what is wrong with the shindo tests. Once I will discover it, I would fix the possible minor bugs.

Ladas:
Heureka,
running this:
export FOG_MOCK=false && bundle exec shindont tests/openstack

it wants credentials in fog/tests/.fog, not ~/.fog for some reason

I'll make sure the tests passes and will send final version

tokengeek:
Bit late to this sorry.

Best way I've found to run shindo test against a real API is:

`FOG_RC=config_file FOG_MOCK=false bundle exec shindont tests/provider`

Which is pretty much the same except declaring the config you want to use rather than relying on that test/.fog that was added.

The real test need to run since the problem is the mock implementation and the real implementation are completely different code paths. There's an issue somewhere for that already.

Basically the mocked code exercises a very small part of fog, ignoring most of the request code.

However you have the tests running now so I'm sure that can be done now :smile:

Ladas:
OK thanks.

So real tests are passing now for Baremetal. 

Although few other tests were not, I am using Juno OpenStack (RDO), so maybe the tests are a bit obsolete? 

Also it can't handle properly missing service, e.g. some special Neutron plugin or Missing Swift. I might look into this later, when there is some time. :-)

Nevertheless this patch should be now ready, or is there something else missing?

Ladas:
@tokengeek @geemus Is there anything else preventing this from merging guys? (except the fog-core dependency I have). It would be great if this could make it in the next release, since I have already 2 patches depending on this.

Thank you guys.

geemus:
Starting to get at least partial green on tests now that fog-core is bumped. Looks like there may be a lingering 1.8 related syntax error at least though, could you take a peek? https://travis-ci.org/fog/fog/jobs/47096971#L897 Thanks!

lib/fog/openstack/requests/baremetal/list_drivers.rb
```diff
@@ -0,0 +1,24 @@
+module Fog
+  module Baremetal
+    class OpenStack
+      class Real
+        def list_drivers
+          request(
+            :expects => [200, 204],
+            :method  => 'GET',
+            :path    => 'drivers',
```
jrafanie: @Ladas I'm guessing the trailing `,` is not 1.8 friendly.

Ladas:
Ah, leftover from my pep8 experience, where trailing commas are recommended. :-)

Ladas:
Hmm there is Fog::Compute test fail. I wonder what I can do about that,

geemus:
Failure looks unrelated, we have a couple intermittent failures...

geemus:
Thanks!

PR #3388, opened 2015-01-10
Add nic_type option

stonith:
Current clone behaviour:
-network_label option not specified - VM NIC type is same as template
-network_label option specified - VM NIC type is hardcoded to VirtualE1000

After nic_type option addition:
-network_label option not specified - VM NIC type is same as template
-network_label option specified - VM NIC type defaults to VirtualE1000 (backwards compatible)
-network_label and nic_type options specified - VM NIC type is what is specified in nic_type option ie. 'nic_type' => 'VirtualVmxnet3'

### DIFF
--- a/lib/fog/vsphere/requests/compute/vm_clone.rb
+++ b/lib/fog/vsphere/requests/compute/vm_clone.rb
@@ -7,6 +7,7 @@ module Fog
           default_options = {
             'force'        => false,
             'linked_clone' => false,
+            'nic_type' => 'VirtualE1000',
           }
           options = default_options.merge(options)
           # Backwards compat for "path" option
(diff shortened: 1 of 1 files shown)


nirvdrum:
I can't readily test this, but the diff looks good and I talked with @stonith about it on IRC.  Just so this is noted somewhere . . . this behavior is already supported when creating or modifying a NIC via the model layer.  However, in that case the attribute name must be :type.  I'm not wild about the varying option names, but in one case it's scoped directly to a NIC and the other is broader, at the VM level.  So the namespacing is appropriate, it's just unfortunate that the options vary based on context.

Thanks again, @stonith, for sticking with this one.

PR #3392, opened 2015-01-13
Remove duplicate openstack_region key

kyletolle:
When running fog in my Rails app under Ruby 2.2.0, I receive the following warning:

`.bundle/ruby/2.2.0/gems/fog-1.27.0/lib/fog/openstack/volume.rb:191: warning: duplicated key at line 196 ignored: :openstack_region`

I removed the duplicate line.


geemus:
Thanks!

kyletolle:
Glad I could help!

PR #3393, opened 2015-01-13
adding option to set ovirt to use filtered API

unorthodoxgeek:
When the ovirt user is not an admin, the client must use the "filtered_api" flag, otherwise all requests fail. Adding the ovirt_filtered_api flag will allow usage of fog with non admin ovirt credentials.


unorthodoxgeek:
that's odd... why are tests failing on jruby 1.9?

geemus:
@unorthodoxgeek the openstack security tests fail intermittently, we suspect a race condition or something along those lines, but haven't ever pinned it down totally. Seems to come and go, which makes the actual debugging painful. I don't think it should be related to anything you changed though.

geemus:
Thanks!

PR #3394, opened 2015-01-13
adding version attribute for ovirt template

unorthodoxgeek:
adding the version attribute for ovirt templates (templates can have the same name, with different versions)

### DIFF
--- a/fog.gemspec
+++ b/fog.gemspec
@@ -78,7 +78,7 @@ Gem::Specification.new do |s|
   s.add_development_dependency('rbvmomi')
   s.add_development_dependency('yard')
   s.add_development_dependency('thor')
-  s.add_development_dependency('rbovirt', '0.0.31')
+  s.add_development_dependency('rbovirt', '0.0.32')
   s.add_development_dependency('shindo', '~> 0.3.4')
   s.add_development_dependency('fission')
   s.add_development_dependency('opennebula')
--- a/lib/fog/ovirt/compute.rb
+++ b/lib/fog/ovirt/compute.rb
@@ -65,6 +65,8 @@ module Fog
             opts[key] = case value
                         when OVIRT::Link
                           value.id
+                        when OVIRT::TemplateVersion
+                          value
                         when Array
                           value
                         when Hash
--- a/lib/fog/ovirt/models/compute/template.rb
+++ b/lib/fog/ovirt/models/compute/template.rb
@@ -19,6 +19,7 @@ module Fog
         attribute :cluster
         attribute :interfaces
         attribute :volumes
+        attribute :version
 
         def interfaces
           attributes[:interfaces] ||= id.nil? ? [] : Fog::Compute::Ovirt::Interfaces.new(


unorthodoxgeek:
this relies on code I've contributed to rvovirt and it's wrong.

geemus:
Just let me know if there is anything around this I can help with. Thanks!

unorthodoxgeek:
I broke the code on another repo and thought the fault was in fog, reopening,

geemus:
@unorthodoxgeek no worries, easy mistake to make. Will we need to wait for some changes in ovirt before we bring this in then?

unorthodoxgeek:
@geemus - maybe I wasn't wrong.
The template version isn't a simple string, it has three attributes, so instead of passing a string I'm now passing an object which has three methods (base_template, version_number and version_name). What I'm getting is the inspect string of that object... ("#<OVIRT::TemplateVersion:0x007f9756088758>"), which is unexpected...
am I doing something wrong?

geemus:
@unorthodoxgeek - that is probably fine actually. It is kind of naive and so I think when it inspects objects it just recursively calls inspect (which would result in what you describe), but that the actual data there would be the object, not the inspect/string. Does that help/make sense?

unorthodoxgeek:
what happens is that template.version returns "#OVIRT::TemplateVersion:0x007f9756088758" instead of the OVIRT::TemplateVersion instance...

geemus:
Ah... hmm. That I am not sure about. I would suspect something about the value being passed in, rather than conversion within the model, but I could be wrong.

unorthodoxgeek:
just to test, I've passed in a hash, that works as expected... I'll investigate further tomorrow :)

geemus:
@unorthodoxgeek cool, yeah that aligns with my expectations. Sounds like a plan, just let me know if there is anything else I can help with. Thanks!

unorthodoxgeek:
eureka. the compute class does to_s, which makes little sense to me, but I guess it makes some sense.
tests will probably fail until rbovirt will get updated, so don't merge this yet

unorthodoxgeek:
ok, this change should work once the rbovirt gem is updated to 0.0.32 - how can we do this?

geemus:
@unorthodoxgeek since not everyone needs rbovirt, I think we depend on the end-user installing it independently. The tests, however, are a different story. You should be able to update the development dependency here: https://github.com/fog/fog/blob/master/fog.gemspec#L81

unorthodoxgeek:
@geemus - updated, a test failed on openstack on one of the runs, I believe it's okay though, isn't it?

geemus:
@unorthodoxgeek yeah, that one fails intermittently for some reason...

Thanks!

unorthodoxgeek:
@geemus - any idea when this will enter a published version? we at foreman are depending on this for a rather important issue...

geemus:
@unorthodoxgeek sorry. I am working on catching up to a bit of issue/PR backlog, but hope to release after that. I'm optimistic that it might be as early as this afternoon, but it might end up being tomorrow. Hopefully by the end of the week regardless.

PR #3397, opened 2015-01-14
Fix small typo

plribeiro3000:
This is a small copy paste fail. This line is not necessary and does not relates to the actual code/deprecation.


geemus:
Thanks!

PR #3400, opened 2015-01-14
Update DigitalOcean docs

petems:
Just some stuff I noticed when using fog + DO for the first time, hopefully it helps! :smile:


geemus:
Thanks!

PR #3403, opened 2015-01-17
Use Fog::Formatador

starbelly:
Don't use Formatador directly


starbelly:
Fails to due to outdated fog-core gem in build

geemus:
Thanks!

PR #3411, opened 2015-01-21
Port Rackspace Orchestration implementation to OpenStack

bzwei:
The goal of this PR is to enhance existing OpenStack Orchestration support to include more information such as resource, events, and template. Since recently added support for Rackspace Orchestration has included all needed features, it is quite natural to port its implementations to OpenStack.

Most of the code for Rackspace can be copied to OpenStack without modification except the namespace. The only enhance I made is to support CFN template format in OpenStack. This is done in model `Template` and `Templates`.

The conflict comes from the existing OpenStack `Stack` and Rackspace `Stack`. The old OpenStack `Stack` class has attributes needed to create a stack, while Rackspace  `Stack` class has attributes associated with a resulting stack. One common attribute named `template` is a raw string in OpenStack `Stack` but a `Template` class in Rackspace `Stack`. It is very hard to merge the old attributes with the new ones without naming conflicts. More importantly it is logically not right to have two kinds of attributes mixed in one class.

I want to keep the implementation of OpenStack Orchestration as close as possible to 
Rackspace Orchestration because we should extract both implementations to a common place in the future. Therefore I override the old OpenStack `Stack` class with the Rackspace one. Other related classes are `Stacks`, and `create_stack`, `update_stack`, `delete_stack`, `list_stack_data` under `requests`. This as a result breaks the backward compatibility. This is the major concern, thus requires attentions and guidances from Fog project collaborators. 

Another matter is the tests. I do not add any tests for OpenStack Orchestration because this work is simply a porting of existing code. Plus I don't see test examples in Rackspace Orchestration that I can follow. These features however are manually tested and examples are given in the `orchestration.md` document.  However, I temporarily commented out tests in `openstack/requests/orchestration/stack_tests` because of above mentioned changes.

### DIFF
--- a/lib/fog/openstack/docs/orchestration.md
+++ b/lib/fog/openstack/docs/orchestration.md
@@ -1,0 +1,295 @@
+# OpenStack Orchestration
+The mission of the OpenStack Orchestration program is to create a human- and machine-accessible service for managing the entire lifecycle of infrastructure and applications within OpenStack clouds.
+
+## Heat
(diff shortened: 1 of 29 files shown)


bzwei:
closed this one and created #3412 because this one reported merge conflicts for no reason.

PR #3412, opened 2015-01-21
Port Rackspace Orchestration implementation to OpenStack

bzwei:
The goal of this PR is to enhance existing OpenStack Orchestration support to include more information such as resources, events, and template. Since recently added support for Rackspace Orchestration has included all needed features, it is quite natural to port its implementations to OpenStack.

Most of the code for Rackspace can be copied to OpenStack without modification except the namespace. The only enhance I made is to support `CFN` template format in OpenStack. This is done in model `Template` and `Templates`.

The conflict comes from the existing OpenStack `Stack` and Rackspace `Stack`. The old OpenStack `Stack` class has attributes needed to create a stack, while Rackspace `Stack` class has attributes associated with a resulting stack. One common attribute named `template` is a raw string in OpenStack `Stack` but a `Template` class in Rackspace `Stack`. It is very hard to merge the old attributes with the new ones without naming conflicts. More importantly it is logically not right to have two kinds of attributes mixed in one class.

I want to keep the implementation of OpenStack Orchestration as close as possible to 
Rackspace Orchestration because we should extract both implementations to a common place in the future. Therefore I override the old OpenStack `Stack` class with the Rackspace one. Other related classes are `Stacks`, and `create_stack`, `update_stack`, `delete_stack`, `list_stack_data` under `requests`. This as a result breaks the backward compatibility. This is the major concern, thus requires attentions and guidances from Fog project collaborators.

Another matter is the tests. I do not add any tests for OpenStack Orchestration because this work is simply a porting of existing code. Plus I don't see test examples in Rackspace Orchestration that I can follow. These features however are manually tested and examples are given in the `orchestration.md `document. However, I temporarily commented out tests in `openstack/requests/orchestration/stack_tests` because of above mentioned changes.


bzwei:
@geemus can you review? Thanks!

geemus:
Yes, although @mdarby may have enough additional context that I should leave it to/for him?

mdarby:
@bzwei yeah, the tests are written in rspec/VCR so I've kept them in a separate repo until we can spin `fog-rackspace` into it's own gem.

It seems as though you've copy/pasted Rackspace Orchestration to OpenStack and renamed classes where necessary. 

I don't see a conflict on my end of things (Rackspace).

bzwei:
@mdarby Yes. Mostly I just copied and pasted, except some changes for template. It now supports both HOT and CFN formats. It also contains the full content of the template, which is needed if we want to redeploy the template.
@geemus and @mdarby This implementation breaks backward compatibility for OpenStack Orchestration. See my first comment. Please instruct.

geemus:
Could we create a shim of some sort for the interim that would check the type of certain attributes and translate if/when needed? I think we have done something to that effect in a few places in the past for the sake of backwards compatibility...

bzwei:
@geemus can you be a little more specific by pointing me to an example? Thanks.

geemus:
Here is a rough example anyway: https://github.com/fog/fog-aws/blob/b5c39ecd093da3917edea6ce7ab88f228649076d/lib/fog/aws/models/compute/snapshots.rb#L19

plribeiro3000:
@bzwei Whats the status on this one? We don't get notified by new commits to we might have lost the track. =s

bzwei:
I committed updates to solve the backward comparability issue. The build failed at OpenStack compute which seems not related to this update. I'll take a closer look.

bzwei:
I have completed the updates and reenabled stack_tests. In summary Rackspace orchestration has been ported to OpenStack. Existing OpenStack orchestration functions still work but are logged as deprecated. 

@geemus, @mdarby, @plribeiro3000 please review.

The Travis CI build has intermittent failures but not due to this PR.

mdarby:
LGTM

geemus:
Awesome, thanks!

kbrock:
Thanks @geemus @bzwei

PR #3414, opened 2015-01-23
Remove the "ovirt_ca_cert_store" option

jhernand:
This option is intended to pass a certificate store containing the
accepted CA certificates, but the underlying "rbovirt" library is about
to remove it because it can only be used with HTTP libraries that
directly support these OpenSSL objects. See here for details:

  https://github.com/abenari/rbovirt/pull/55

As an alternative to this "ovirt_ca_cert_store" the already existing
"ovirt_ca_cert_file" can be used, with the same functinality. Actually
the "ovirt_ca_cert_store" parameter isn't currently used in the fog code

Signed-off-by: Juan Hernandez <juan.hernandez@redhat.com>

### DIFF
--- a/lib/fog/ovirt/compute.rb
+++ b/lib/fog/ovirt/compute.rb
@@ -5,7 +5,7 @@ module Fog
     class Ovirt < Fog::Service
       requires   :ovirt_username, :ovirt_password
       recognizes :ovirt_url,      :ovirt_server,  :ovirt_port, :ovirt_api_path, :ovirt_datacenter,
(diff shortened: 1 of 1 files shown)


geemus:
Thanks. I figured I would wait for now until this lands in rbovirt, if I don't catch that happening feel free to ping me and I'll try to finish tying this all together. Thanks!

jhernand:
Great, thanks. I will let you now when the change is merged in rbovirt.

jhernand:
This change in rbovirt won't happe, so I'm closing this pull request.

geemus:
Too bad, thanks for the update though.

PR #3415, opened 2015-01-23
Fixes "Error - undefined method 'delete' for nil:NilClass" when attempting to stop a container

witlessbird:
...ng to stop a container


geemus:
Looks good, thanks!

PR #3417, opened 2015-01-26
[google] Add support for JSON key format

### LINKED ISSUE #3413: [google] Add support for JSON key format
Google introduced a new Service Account key format (JSON) and over time this will likely become the preferred format.  We should begin by supporting this new format soon. If Google does make this the default, we may want to follow-up and issue a (debug | info) warning to users relying on P12 format.

AFAIK, there is no clear documentation on this format. However, if you open up the Developers Console and look under "APIs & Auth" -> "Credentials", you'll see the "Generate New JSON Key" button next to any existing Service Accounts (or be given the option to create one with a new SA).

/cc @icco, @frodenas

frodenas:
This fixes #3413.

@erjohnso, @icco I'm not sure if P12 format is deprecated, so I didn't add any *warning* message. We can add it later if it's finally deprecated.

### DIFF
--- a/lib/fog/google/compute.rb
+++ b/lib/fog/google/compute.rb
@@ -4,7 +4,8 @@ module Fog
   module Compute
     class Google < Fog::Service
       requires :google_project
-      recognizes :app_name, :app_version, :google_client_email, :google_key_location, :google_key_string, :google_client
+      recognizes :app_name, :app_version, :google_client_email, :google_key_location, :google_key_string,
+                 :google_client, :google_json_key_location, :google_json_key_string
 
       GOOGLE_COMPUTE_API_VERSION     = 'v1'
       GOOGLE_COMPUTE_BASE_URL        = 'https://www.googleapis.com/compute/'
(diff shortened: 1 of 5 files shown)


icco:
Would you be interested in adding some tests? This is quickly becoming a relatively complicated set of options...

erjohnso:
@frodenas - thanks for this!!

P12 is not deprecated yet, so it's totally fine to not warn now.  However, it would probably be a good idea to update the docs (http://fog.io/compute/?#google-compute-engine). I'd suggest updating the code-snippet to use the JSON format, but mention in text that the P12 format is also supported.

As @icco said, it's getting a bit complicated, so better docs on the options will help users too.

frodenas:
Agreed, I'll work on tests and docs hopefully this weekend.

erjohnso:
Thanks Ferdy!

icco:
(I can review docs as well. Docs are stored in https://github.com/fog/fog.github.com)

geemus:
Thanks!

PR #3423, opened 2015-02-01
Support for DigitalOcean user_data when creating server

korczis:


korczis:
Does not works with v1, closing.

geemus:
Thanks, let me know if there is anything here I can help with.

korczis:
@geemus: It will be nice if we will be able to engage v2 API, I will be glad to help.

Officially it still public beta but it seems that even official web ui depends on v2 api ...

PR #3424, opened 2015-02-02
[HP|storage] Add support for headers to get_object

bruz:
This allows specifying a Range header, which in turn results in a 206 response status.


geemus:
Thanks!

PR #1978, opened 2013-07-24
[Vsphere] added list/get compute resources functions

mingjin:
This patch is to add list/get compute resources of vsphere functions to fog. 

Usage:
to list all compute resources: compute.list_compute_resources
to just list effective compute resources(poweron + connected + not in maintenance state) : compute.list_compute_resources({:effective => true})
to get compute resources by name: compute.get_compute_resource('xxx.xxx.xxx.xxx')

Notice:
Vsphere abstracts clusters and hosts that don't belong to any cluster evenly as compute resource, so the return result may refer to either cluster or single host. If you want to get only hosts or clusters, you can use attribute 'isSingleHost' to determine the type.

PR #3426, opened 2015-02-03
[vSphere] List compute resources, orginally PR#1978

mingjin:
@nirvdrum 

I created a new feature branch to filter out the unrelated commits. Please let me know any question.

Descriptions quoted from #1978:
This patch is to add list/get compute resources of vsphere functions to fog.

Usage:
to list all compute resources: compute.list_compute_resources
to just list effective compute resources(poweron + connected + not in maintenance state) : compute.list_compute_resources({:effective => true})
to get compute resources by name: compute.get_compute_resource('xxx.xxx.xxx.xxx')
to determine if the returned compute resource is single host, check the attribute 'isSingleHost'

Notice:
Vsphere abstracts clusters and hosts that don't belong to any cluster evenly as compute resource, so the return result may refer to either cluster or single host. If you want to get only hosts or clusters, you can use attribute 'isSingleHost' to determine the type.

### DIFF
--- a/lib/fog/vsphere/requests/compute/get_compute_resource.rb
+++ b/lib/fog/vsphere/requests/compute/get_compute_resource.rb
@@ -17,6 +17,23 @@ module Fog
 
       class Mock
         def get_compute_resource(name, datacenter_name)
+          {
+            :id=>"domain-s7", 
+            :name=>"fake-host", 
+            :totalCpu=>33504, 
+            :totalMemory=>154604142592, 
+            :numCpuCores=>12, 
+            :numCpuThreads=>24, 
+            :effectiveCpu=>32247, 
+            :effectiveMemory=>135733, 
+            :numHosts=>1, 
+            :numEffectiveHosts=>1, 
+            :overallStatus=>"gray", 
+            :overallCpuUsage=>15682, 
+            :overallMemoryUsage=>132755, 
+            :effective=>true, 
+            :isSingleHost=>true
+          }
         end
       end
     end
--- a/lib/fog/vsphere/requests/compute/get_compute_resource.rb
+++ b/lib/fog/vsphere/requests/compute/get_compute_resource.rb
@@ -1,0 +1,24 @@
+module Fog
+  module Compute
+    class Vsphere
+      class Real
+        def get_compute_resource(name, datacenter_name)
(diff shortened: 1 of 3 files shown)


lib/fog/vsphere/requests/compute/get_compute_resource.rb
```diff
@@ -0,0 +1,24 @@
...
+        protected
+
+        def get_raw_compute_resource(name, datacenter_name)
+          find_raw_datacenter(datacenter_name).find_compute_resource(name)
+        end
+      end
+
+      class Mock
+        def get_compute_resource(name, datacenter_name)
+        end
```
nirvdrum: Is there a hash or something we can return as an appropriate mocked response?

lib/fog/vsphere/requests/compute/list_compute_resources.rb
```diff
@@ -0,0 +1,58 @@
+module Fog
+  module Compute
+    class Vsphere
+      class Real
+        def list_compute_resources(filters = { })
```
nirvdrum: I need to double-check, but I think the other vSphere hashes are returned with String keys.  If that's the case, I recommend changing here for the sake of consistency.

mingjin: @nirvdrum, understand, we checked the implementations like list_vm_volumes/list_datastores, they all return symbol keys.

lib/fog/vsphere/requests/compute/get_compute_resource.rb
mingjin: @nirvdrum that's a good idea. I'll add appropriate mocked responses tomorrow.

mingjin:
@nirvdrum I've added appropriate mock responses for list_compute_resources, please let me know if any question.

nirvdrum:
Thanks!  It'd be great to get a corresponding model class for these requests, but having the requests in place is a great start.

mingjin:
@nirvdrum thanks. We'll do better in the future pull requests.

nirvdrum:
No worries.  Thanks again.  Especially taking the time to clean up the PR -- that made it much easier to merge.

geemus:
Thanks!

Issue #3432, opened 2015-02-10
Implement floating disks attach and detach action in rbovirt provider


pdilung:
Implement disk attach and detach action in _ovirt_ provider based on abenari/rbovirt#60 proposal and abenari/rbovirt#61 pull request.

pdilung:
I have made an initial implementation of the feature. Please see 6b64bd08cab234292034b94f000f620801759a52 for details.
Please note that this feature must be supported by `rbovirt` library as well. Currently, there is a pending pull request abenari/rbovirt#61. Once it is merged, I shall open a pull request on fog part as well.

geemus:
Sounds good, thanks!

pdilung:
Created a pull request #3435

geemus:
Thanks! I'll close this in favor of continuing discussion on the pull as needed.

PR #3435, opened 2015-02-13
[rbovirt]  Implement floating disks attach and detach action

pdilung:
As described in #3432


plribeiro3000:
@pdilung Could you add some tests around so we can make sure we won't break it in the future?

lib/fog/ovirt/requests/compute/attach_volume.rb
```diff
@@ -0,0 +1,22 @@
+module Fog
+  module Compute
+    class Ovirt
+      class Real
+        def attach_volume(id, options)
+          raise ArgumentError, "instance id is a required parameter" unless id
+          raise ArgumentError, "volume id is a required parameter for destroy-volume" unless options.key? :id
```
abenari: Error message says destroy instead of attach

pdilung: Fixed in 441d90111a285fc74f756c3bfcf5ec3e7e128895

pdilung:
@plribeiro3000 I would love to do this together with pull request. Unfortunately, I am still a Ruby rookie and haven't found tests for neither for volumes collection nor for add_volume/destroy_volume I could use as a starting point. I would appreciate any help with writing tests for this feature. Anyone could help? Thanks

abenari:
@pdilung re help with tests:
Add the new request [here](https://github.com/fog/fog/blob/master/tests/ovirt/compute_tests.rb#L18)
Add two new test for attach and detach request similar to the [update volume tests](https://github.com/fog/fog/blob/master/tests/ovirt/requests/compute/update_volume_tests.rb)

pdilung:
@abenari, @plribeiro3000 I may give it a try but shindo and tests are still fog for me :) I will ask someone to write them in case I get stuck.
@abenari :+1: for you constant support

pdilung:
@plribeiro3000, @geemus, @abenari Any chance to get this into upstream without tests? I would need someone's help to write them. Thanks a lot.

plribeiro3000:
The changes are pretty simple but i think  @geemus is the best to answer that. I will take another look at the code and see if i can catch something.

lib/fog/ovirt/models/compute/server.rb
```diff
@@ -89,6 +89,16 @@ def update_volume attrs
           service.update_volume(id, attrs)
         end
 
+        def attach_volume attrs
```
plribeiro3000: Can you add `(` and `)` to follow our conventions?

lib/fog/ovirt/models/compute/server.rb
```diff
@@ -89,6 +89,16 @@ def update_volume attrs
...
+          wait_for { stopped? } if attrs[:blocking]
+          service.attach_volume(id, attrs)
+        end
+
+        def detach_volume attrs
```
plribeiro3000: `(` and `)` here too.

lib/fog/ovirt/requests/compute/attach_volume.rb
```diff
@@ -0,0 +1,22 @@
+module Fog
+  module Compute
+    class Ovirt
+      class Real
+        def attach_volume(id, options)
+          raise ArgumentError, "instance id is a required parameter" unless id
+          raise ArgumentError, "volume id is a required parameter for attach-volume" unless options.key? :id
+
+          client.attach_volume(id, options[:id])
```
plribeiro3000: Why would you pass a hash (`options`) to use only one of its attributes (`options[:id]`) ?

plribeiro3000: Perhaps `def attach_volume(instance_id, volume_id)` is a better idea.

pdilung: I agree, the reason I did this is that all `*_volume` methods use same arguments. For instance [`destroy_volume`](https://github.com/fog/fog/blob/master/lib/fog/ovirt/requests/compute/destroy_volume.rb) would only need a `volume_id` to be removed.
Should I update?

plribeiro3000:
I didn't get the point with the exceptions. Are you trying to check for nil values?

lib/fog/ovirt/requests/compute/attach_volume.rb
plribeiro3000: Nope. Let as is for now. Its a strong point to be refactored later i guess.

plribeiro3000:
With your last comment i just realized the current api is doing that so i think it is ok to let as is for now.

@geemus Something else to add here?

pdilung:
I try to make sure that volume id is passed to a method before I call `client.attach_volume(id, options[:id])`. See [`destroy_volume`](https://github.com/fog/fog/blob/master/lib/fog/ovirt/requests/compute/destroy_volume.rb) for reference.

lib/fog/ovirt/requests/compute/attach_volume.rb
pdilung: Agree :+1:

pdilung: Should I also revert back and remove braces around `attrs` from `Fog::Compute::Ovirt::Server#attach_volume` and `Fog::Compute::Ovirt::Server#detach_volume` methods definition ? Personally, I would keep the braces.

plribeiro3000: Nope. Keep the braces. It is a style change so nothing breaks.

The api is one to concern because all apis are like that the user would expect that to be the same so we can't change just this one. Also, this refactoring is out of the scope of this PR.

Feel free to create another PR to change this. =)

geemus:
Thanks! Getting some tests added soon would be great also.

pdilung:
Ok, I try my bests but might need some guidance.

geemus:
Sure thing, just let us know as questions arise and we'll do what we can to
help out. Thanks!

PR #3438, opened 2015-02-14
vm_clone used regular virtual switches no matter what type of portgroup ...

sofam:
...was named, changed it to look up and handle distributed virtual switches correctly. @nirvdrum

### DIFF
--- a/lib/fog/vsphere/requests/compute/vm_clone.rb
+++ b/lib/fog/vsphere/requests/compute/vm_clone.rb
@@ -116,11 +116,21 @@ module Fog
           # Options['network']
           # Build up the config spec
           if ( options.key?('network_label') )
-            #network_obj = datacenter_obj.networkFolder.find(options['network_label'])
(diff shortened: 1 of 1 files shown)


sofam:
@nirvdrum any comments on this?

nirvdrum:
Sorry.  I forgot about this.  Just to be clear, this is a problem you're seeing even with the merged fix from #3121 applied?

sofam:
@nirvdrum it's actually unrelated (kind of) because the problem lies in the port backing chosen in vm_clone, before my pull request vm_clone would just create a VirtualEthernetCardNetworkBackingInfo port backing, no matter what kind of network was specified, my pull request checks the type of the network chosen, if it's a normal vSwitch it uses VirtualEthernetCardNetworkBackingInfo, and if it's a distributed port group it uses VirtualEthernetCardDistributedVirtualPortBackingInfo.

nirvdrum:
This looks good.  The code formatting doesn't match the new guidelines, but I can clean that up easily enough.  Thanks!

geemus:
Thanks!

PR #3443, opened 2015-02-18
[aws|compute] adding support for c4 instance class

ddoc:


plribeiro3000:
Hmmmm. Your fork is outdated. This file does not exist in the main `Fog` repo anymore.

The necessary change is in `fog-aws` which was already merged. So this can be safely closed.

Thanks!

PR #3444, opened 2015-02-18
Remove duplicate lines from code example

spajus:
If `put_object_manifest` has to be called 3 times, it is probably a problem :)


geemus:
Good catch, thanks!

PR #3446, opened 2015-02-19
Use fog-core 1.27.4 and above

mattbostock:
fog-core version 1.27.4 includes this bug fix:
https://github.com/fog/fog-core/commit/58556e4b1e65d2ac8dd177c531dd8a23929d39e3

Without that fix, Fog fails with:

    NoMethodError:
           undefined method `redisplay_progressbar' for Fog::Formatador:Class

### DIFF
--- a/fog.gemspec
+++ b/fog.gemspec
@@ -46,7 +46,7 @@ Gem::Specification.new do |s|
 
   ## List your runtime dependencies here. Runtime dependencies are those
   ## that are needed for an end user to actually USE your code.
-  s.add_dependency("fog-core", "~> 1.27", ">= 1.27.3")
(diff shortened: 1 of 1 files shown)


mattbostock:
The tests fail for jruby-19 only, though I'm not sure why. Are you able to provide any clues?

plribeiro3000:
I'm afraid this might lead to several conflicts since the providers are depending on specific versions of `fog-core` as well.

Perhaps you could reference `fog-core` in your `Gemfile` instead? Can you try this solutions instead?

plribeiro3000:
hmm. It was already pinned to `1.27.3` so this is not a big change.

Thanks @mattbostock !

geemus:
Thanks!

PR #3449, opened 2015-02-20
Allow setting gateway option in CustomizationIPSettings

jipjipme:
The "gateway" field of CustomizationIPSettings (http://pubs.vmware.com/vsphere-55/topic/com.vmware.wssdk.apiref.doc/vim.vm.customization.IPSettings.html) is mentioned in vm_clone but oddly never used, and causes an error if specified.

Tested against vSphere 5.5.

### DIFF
--- a/lib/fog/vsphere/requests/compute/vm_clone.rb
+++ b/lib/fog/vsphere/requests/compute/vm_clone.rb
@@ -161,8 +161,11 @@ module Fog
             end
             raise ArgumentError, "domain is required" unless cust_options.key?("domain")
             cust_domain = cust_options['domain']
-            cust_ip_settings = RbVmomi::VIM::CustomizationIPSettings.new(cust_options["ipsettings"]) if cust_options.key?("ipsettings")
-            cust_ip_settings.ip = RbVmomi::VIM::CustomizationFixedIp("ipAddress" => cust_options["ipsettings"]["ip"]) if cust_options.key?("ipsettings")
+            if cust_options.key?("ipsettings")
+              cust_ip_settings = RbVmomi::VIM::CustomizationIPSettings.new(cust_options["ipsettings"])
+              cust_ip_settings.ip = RbVmomi::VIM::CustomizationFixedIp("ipAddress" => cust_options["ipsettings"]["ip"])
+              cust_ip_settings.gateway = cust_options['ipsettings']['gateway'] if cust_options['ipsettings'].key?('gateway')
+            end
             cust_ip_settings ||= RbVmomi::VIM::CustomizationIPSettings.new("ip" => RbVmomi::VIM::CustomizationDhcpIpGenerator.new())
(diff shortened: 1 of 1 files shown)


geemus:
@nirvdrum could you review? Thanks!

nirvdrum:
I'll take a look.  There are generally a lot of problems and limitations of dealing with NICs during clone.  I recommend just clone, blow away the NICs, and then create what you need in the configuration you want.  But if this is simple enough to support, I'm fine merging it.  I'll try it out on a vSphere cluster this weekend.

jipjipme:
Much appreciated. I had a bunch of pre-existing templates with one NIC defined, and this let me do everything in one shot during the clone.

nirvdrum:
The problems are:

* vm_clone needs to effectively duplicate logic from the NIC creation process and routinely seems to fall short in that department
* The vSphere API limits what you can do during a clone (e.g., you can't change the number of NICs)

But it sounds like your use case falls into what vSphere expects so we should support it.

FYI, you can see how I handle it here:

https://github.com/rubber/rubber/blob/853fba652e09cde952153aaea53cfe1d859aa1c5/lib/rubber/cloud/vsphere.rb#L32-L99

Basically I clone, destroy the NICs, then recreate them with a new customization spec.  It works rather nicely and works in all known cases.

lib/fog/vsphere/requests/compute/vm_clone.rb
```diff
@@ -161,8 +161,11 @@ def vm_clone(options = {})
             end
             raise ArgumentError, "domain is required" unless cust_options.key?("domain")
             cust_domain = cust_options['domain']
-            cust_ip_settings = RbVmomi::VIM::CustomizationIPSettings.new(cust_options["ipsettings"]) if cust_options.key?("ipsettings")
-            cust_ip_settings.ip = RbVmomi::VIM::CustomizationFixedIp("ipAddress" => cust_options["ipsettings"]["ip"]) if cust_options.key?("ipsettings")
+            if cust_options.key?("ipsettings")
+              cust_ip_settings = RbVmomi::VIM::CustomizationIPSettings.new(cust_options["ipsettings"])
+              cust_ip_settings.ip = RbVmomi::VIM::CustomizationFixedIp("ipAddress" => cust_options["ipsettings"]["ip"])
+              cust_ip_settings.gateway = cust_options['ipsettings']['gateway'] if cust_options['ipsettings'].key?('gateway')
```
nirvdrum: What happens if you set `cust_ip_settings.gateway` to `nil`?  I suspect it's a safe operation and that the default is already `nil`.  If that's the case, we can remove the `if` statement modifier and just always assign.

jipjipme: Yep, that works just fine.

nirvdrum:
Thanks!

geemus:
Thanks!

PR #3452, opened 2015-02-22
Extract local storage provider

### LINKED ISSUE #3439: Extract local provider to a separate gem
I tried to use extracted aws provider without the full fog gem as dependency. This worked, but I also needed the local provider for development environment. So, I went ahead and extracted it out as [a separate repo](https://github.com/lautis/fog-local). Would you be interested having fog-local as a separate gem?

There's still some things to do, such as

* moving the repository under fog organisation
* publishing the gem
* replacing local provider in fog with the extracted gem as a dependency

lautis:
Use `fog-local` gem to implement local storage. Removes all local storage related code from this repo. The extracted gem is published to rubygems.org and has fog owners listed as gem owners.

The tests for the local provider are also removed, but the bin spec is left untouched. Testing bin seems a bit harder in extracted gems as it requires the full fog gem and uses different test runner. This is the same pattern as with RiakCS in #3425.

Closes #3439.


geemus:
Thanks!

PR #3454, opened 2015-02-23
[ovirt] Implement alias attribute for Fog::Compute::Ovirt::Volume class

pdilung:
Implement `alias` attribute for `Fog::Compute::Ovirt::Volume` class according to https://github.com/abenari/rbovirt/pull/66.


geemus:
Thanks!

PR #3455, opened 2015-02-24
[ovirt] Implement affinity groups

pdilung:
Implement affinity groups into _ovirt_ provider as described in https://github.com/abenari/rbovirt/issues/67 and https://github.com/abenari/rbovirt/pull/69.


geemus:
Could we add some testing around this? I'm afraid otherwise it may be fragile and/or prone to future breakage.

pdilung:
@geemus Yes, that is something I would like to do soon. As I wrote in #3435 I would need some help and time for this. Going to ask for it in next two weeks.
I would much appreciate if you could merge the PR before and I create tests PRs for #3435 and for this one as well :-)

geemus:
Sounds like a plan, just let me know how I can help.

pdilung:
Sure, thanks a lot.

PR #3456, opened 2015-02-24
Use head version of all supported, released MRI versions

calebthompson:


.travis.yml
```diff
@@ -7,11 +7,10 @@ script: bundle exec rake travis
 matrix:
   fast_finish: true
   include:
-    - rvm: 1.9.3
```
plribeiro3000: Could you not remove `1.9.3` yet?

.travis.yml
```diff
@@ -7,11 +7,10 @@ script: bundle exec rake travis
...
-    - rvm: 2.0.0
-    - rvm: 2.1.0
-    - rvm: 2.1.5
-    - rvm: 2.1.5
+    - rvm: 2.0
+    - rvm: 2.1
+    - rvm: 2.2
+    - rvm: 2.2
```
plribeiro3000: You have `2.2` twice. I know it was wrong before but perhaps you can fix it as well?

calebthompson: I believe this second one is to run on 2.2 with the head versions of gems, whereas the previous 2.2 uses the released versions.

.travis.yml
calebthompson: Are you sure? https://www.ruby-lang.org/en/news/2015/02/23/support-for-ruby-1-9-3-has-ended/

fog.gemspec
```diff
@@ -45,7 +45,7 @@ Gem::Specification.new do |s|
   s.extra_rdoc_files = %w[README.md]
 
   ## Ruby version
-  s.required_ruby_version = ">= 1.9.3"
+  s.required_ruby_version = ">= 2"
```
plribeiro3000: We didn't get to a decision about dropping `1.9.3` so perhaps we need to know if for sure first.

.travis.yml
plribeiro3000: But it does not make sense, does it?

.travis.yml
plribeiro3000: We didn't decide about `1.9.3` yet. Perhaps the end of support might change this scenario.

@tokengeek @geemus Thoughts?

.travis.yml
calebthompson: I think it does. We're doing a separate build of our changes in `fog` against released versions of `fog-core` and `fog-json` and also against the master branches of those two gems.

plribeiro3000: But you don't need to list it twice to play with the matrix, do you?

calebthompson: I'm not entirely sure. It appears that master has duplicate entries as well. https://github.com/fog/fog/blob/master/.travis.yml#L22-L25

This is how the Travis docs recommend, but we end up with several more runs (each version against edge), which could be good:

```yml
rvm:
  - 2.0
  - 2.1
  - 2.2
  - jruby-19mode
  - jruby-head

gemfile:
  - Gemfile
  - gemfiles/Gemfile-edge

matrix:
  fast_finish: true

  allow_failures:
    - rvm: jruby-head
```

calebthompson: http://docs.travis-ci.com/user/build-configuration/

calebthompson: Do you want that instead?

plribeiro3000: Yeah. I think that the duplication wrong and is not needed.

plribeiro3000: Awesome. This way it is more clear and should be the same as before. :+1:

fog.gemspec
geemus: I think we should probably leave this as 1.9.3 for now, though dropping it also in time would be good.

calebthompson:
1.9.3 support re-added.

geemus:
Thanks!

plribeiro3000:
@calebthompson Could you add `2.2` as an allowed failure for now?

lib/fog.rb
```diff
@@ -23,6 +23,7 @@
 require 'fog/aws'
 require 'fog/bluebox'
 require 'fog/brightbox'
+require 'fog/cloudfoundry'
```
tokengeek: This needs removing since it's unrelated to the PR and Travis can't load it.

tokengeek:
Re: 1.9.3 support - this is still in place for the foreseeable future because lots of linux distros in production seem to have 1.9.3 as the default.

Obviously this branch/work is for a major change but when we discussed it (https://github.com/fog/fog/pull/3348#issuecomment-67451436) that was raised as a issue.

I need to catch @geemus and @krames (maybe @plribeiro3000 ) and finalise a minimal "roadmap" (not scattered across dozens of tickets) with a few "major" releases over the next few months. Then 1.9.3 can go with 3.0 or 4.0. 2.0 being a minimal drop 1.8.7 release.

Unfortunately fog 2.0 has become the branding for the "next generation" of fog which was a mistake on my part.

plribeiro3000:
@calebthompson Could you take care of @tokengeek comment?
I think it is the last point to get this merged. :+1:

calebthompson:
Done.

plribeiro3000:
Thanks @calebthompson!

geemus:
Thanks!

PR #3457, opened 2015-02-24
Adds `ovirt_filtered_api` to list of recognized options for Ovirt service

hallettj:
The `ovirt_filtered_api` option instructs rbovirt to use oVirt's non-admin API, so that regular users can make API calls.


geemus:
Thanks!

PR #3460, opened 2015-02-26
Expanded tests

tokengeek:
An assortment of specs and test changes I found on old branches.

Support work for existing behaviours to support refactoring.

### DIFF
--- a/tests/helper.rb
+++ b/tests/helper.rb
@@ -15,7 +15,6 @@ if Fog.mocking?
   Fog::Logger.warning "Setting default fog timeout to #{Fog.timeout} seconds"
 
   # These sets of tests do not behave nicely when running mocked tests
-  Thread.current[:tags] << '-serverlove'
   Thread.current[:tags] << '-xenserver'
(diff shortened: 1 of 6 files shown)


tests/helper.rb
```diff
@@ -13,6 +13,12 @@
   FOG_TESTING_TIMEOUT = ENV['FOG_TEST_TIMEOUT'] || 2000
   Fog.timeout = 2000
   Fog::Logger.warning "Setting default fog timeout to #{Fog.timeout} seconds"
+
+  # These sets of tests do not behave nicely when running mocked tests
+  Thread.current[:tags] << '-serverlove'
```
plribeiro3000: Is there any serverlove code left here?

tokengeek: It looks like since I originally wrote this the tests were removed. Which means this isn't much use.

I'll update accordingly.

plribeiro3000:
:+1: @tokengeek

geemus:
Thanks!

Issue #3459, opened 2015-02-25
Joyent Connection SSL Verify Error


hulu1522:
Connecting to a Joyent Cloudapi service fails if using a self-signed certificate. (Note: using test-kitchen also)

```
>>>>>> Create failed on instance <default-ubuntu-1404>.
>>>>>> Please see .kitchen/logs/default-ubuntu-1404.log for more details
>>>>>> ------Exception-------
>>>>>> Class: Kitchen::ActionFailed
>>>>>> Message: Unable to verify certificate, please set `Excon.defaults[:ssl_ca_path] = path_to_certs`, 
`ENV['SSL_CERT_DIR'] = path_to_certs`, `Excon.defaults[:ssl_ca_file] = path_to_file`, 
`ENV['SSL_CERT_FILE'] = path_to_file`, `Excon.defaults[:ssl_verify_callback] = callback` (see OpenSSL::SSL::SSLContext#verify_callback),
 or `Excon.defaults[:ssl_verify_peer] = false` (less secure).
>>>>>> ----------------------
```

Please allow an option to be passed in that will disable this.  The only fix I found was to add the following to [this](https://github.com/fog/fog/blob/master/lib/fog/joyent/compute.rb) file:
```
Excon.defaults[:ssl_verify_peer]=false
```

This was also suggested as a fix for a different issue [here](https://github.com/joyent/knife-joyent/issues/6#issuecomment-5267278).

geemus:
To clarify, where/how do you imagine the option would be passed in? Thanks!

hulu1522:
@geemus  I don't know if this is proper implementation but something similar to this would work:

https://github.com/hulu1522/fog/commit/25db88412a8e7b3ba56fcd58b4e55c083409a0c1

geemus:
Closing in favor of further discussion on #3464

PR #3464, opened 2015-02-27
allow self-signed certs for Joyent

hulu1522:


hulu1522:
Allows a fix to issue #3459

lib/fog/joyent/compute.rb
```diff
@@ -116,6 +118,10 @@ class Real
         def initialize(options = {})
           @connection_options = options[:connection_options] || {}
           @persistent = options[:persistent] || false
+          
+          if options[:joyent_ssl_verify_peer].to_s.downcase == 'false'
+            Excon.defaults[:ssl_verify_peer]=false
```
geemus: This would "leak" out to other excon usage, which seems bad. I think if you merge it in to the connection options above instead it would properly scope this to just this case. In fact, you could probably just pass in `:connection_options => { :ssl_verify_peer => false }` and get it to work without code modification. Should be worth a try anyway, thanks!

hulu1522:
@geemus   Thanks for the suggestion.  I went ahead and updated the [kitchen-joyent](https://github.com/hulu1522/kitchen-joyent/commit/2c7d1c1c5b58e527e0afd30c8011a0769a3d61aa) PR to include this option.  If you think it is a better solution then we should just close this PR.

geemus:
Yeah, I think at least for the time being that would be my preference. Sorry I didn't think about/suggest it sooner! Thanks.

geemus:
And then I clicked the wrong button! Oh well...

hulu1522:
Well you could revert it or just fix it on your next release.  The kitchen-joyent PR that I have waiting doesn't touch these anymore so we should be good.  Thanks for the help.  :)

geemus:
Yeah, I think it should be reverted now.

PR #3468, opened 2015-03-04
Fix broken VDI creation by doing a partial rollback

kongslund:
The previous change to create_vdi.rb breaks existing functionality even though it is supposed to be backwards compatible. This pull request reverts the change back to the previous working state.

## How does it break?

The condition `if config[:storage_repository].nil? || config[:SR].nil?` means that both `:storage_repository` _and_ `:SR` must be defined which is one too many.

When specifying `:SR` as part of a `compute.vdis.create :name => 'xxx', :SR => sr.reference, ...` it only appears as `:__sr` in `config` when inside the `create_vdi` method. I guess this is due to aliasing but it means that `:SR` is never defined in `config` and is thus always missing.

The `present?` method is not available in my runtime environment. It seems to be provided by Rails?

Ping: @plribeiro3000

### DIFF
--- a/lib/fog/xenserver/requests/compute/create_vdi.rb
+++ b/lib/fog/xenserver/requests/compute/create_vdi.rb
@@ -3,35 +3,16 @@ module Fog
     class XenServer
       class Real
         def create_vdi( config )
-          raise ArgumentError.new('Invalid #config') if config.nil?
-          raise ArgumentError.new('Missing #virtual_size attribute') if config[:virtual_size].nil?
-          raise ArgumentError.new('Missing #read_only attribute') if config[:read_only].nil?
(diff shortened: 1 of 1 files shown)


plribeiro3000:
Ouch. @kongslund Thank you for the feedback. Can you try `fog-xenserver` and see if it works as intended?

This deprecation is intended to be floating around for some time until we cut it off from `fog` and add the dependency to `fog-xenserver` that is much more maintainable.

In the meanwhile i will try to figure out how to keep both states.

What do you think? Can we work this out?

plribeiro3000:
@kongslund Could you take a look at #3469? Thanks

plribeiro3000:
I'm closing this in favor of #3469. Thanks!

PR #3469, opened 2015-03-04
[Xenserver] Fix Xenserver VDI creation

plribeiro3000:
When adding deprecations in https://github.com/fog/fog/pull/3357 the request to create VDI was broked.
This changes fixes the bad logic and add support for both cases (old and new api)
which means it is backwards compatible.

See https://github.com/fog/fog/pull/3468 for more info

cc/ @kongslund


plribeiro3000:
@kongslund i will go ahead and merge this. Let me know if it works for you.

kongslund:
Great, I will try it out tomorrow!

geemus:
Thanks!

kongslund:
When providing `:storage_repository` it works.
When providing either `:SR` or `:__sr` it fails.

I believe that `:__sr` should be a String representing an opaque reference instead of being treated as a model object of type StorageRepository.

I've created a new PR (#3471) in which it works when just specifying `:SR` as such:

    vdi = conn.vdis.create :name_label => 'my-disk',
                           :SR => sr.reference
which I believe is also the idea with the interface in fog-xenserver.

PR #3471, opened 2015-03-05
[Xenserver] Fix Xenserver VDI creation, take two

kongslund:
Treat `:__sr` as a String representing an opaque reference instead of treating it as a model object of type StorageRepository.

So now when invoking `compute.vdis.create` you can either provide :storage_repository or :SR.

    # The new way
    vdi = compute.vdis.create :name_label => 'my-disk',
                           :SR => sr.reference
    
    # The deprecated way
    vdi = compute.vdis.create :name_label => 'my-disk',
                           :storage_repository => sr

See the discussions in #3469 and #3468.

cc @plribeiro3000


plribeiro3000:
Good catch @kongslund! Thanks!

geemus:
Thanks!

PR #3473, opened 2015-03-05
zone.id instead of zone.domain for record.destroy

hh:
domain=nil for me... but id looks good

```
dme.zones.find{ |z| z.id == 'foo.instantinfrastructure.com'}.records.find_all{|r| r.name=='foo' && r.type=='A'}.first.zone
=>   <Fog::DNS::DNSMadeEasy::Zone
    id="foo.instantinfrastructure.com",
    domain=nil,
    gtd_enabled=nil,
    nameservers=nil
  >
```

### DIFF
--- a/lib/fog/dnsmadeeasy/models/dns/record.rb
+++ b/lib/fog/dnsmadeeasy/models/dns/record.rb
@@ -28,7 +28,7 @@ module Fog
         end
 
         def destroy
-          service.delete_record(zone.domain, identity)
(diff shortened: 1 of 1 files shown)


geemus:
Looks good, ids are almost always the better idea anyway (if you have them) since they won't change but other values sometimes will. Not totally sure why it has nil domain in this context, but seems like the right thing to do anyway. Thanks!

PR #3474, opened 2015-03-06
[GleSYS] Add ability to persist changes on existing server

Artmann:
Will make it possible to save chages to servers.

    server = compute.servers.get(serverid)
    server.memorysize = 1024
    server.cpucores = 1
    server.save


geemus:
Awesome, thanks!

PR #3475, opened 2015-03-06
[ovirt] Fix update_interface method

pdilung:
It seems that uderlying `OVIRT::Client#update_interface` method has evolved and the parameters of this method have been changed. Because of this, an exception is thrown whenever `Fog::Compute::Ovirt::Real#update_interface` is called. This PR amends this.


geemus:
Thanks!

PR #3476, opened 2015-03-09
Rackspace Neutron Security Groups & Rules

mdarby:
This pull adds SecurityGroup and SecurityGroupRules for Neutron at Rackspace.
http://developer.openstack.org/api-ref-networking-v2.html

Tests exist in an external repo and will be merged in.
fog-test 2015-03-09 16-33-48


geemus:
Thanks!

PR #3485, opened 2015-03-15
Add spec_helper and coverage

tokengeek:
To help DRY out specs and making it easier to introduce coverage
this rejiggs the specs to use a simple `spec_helper` rather than
all having repeated requires.


geemus:
Thanks!

PR #3489, opened 2015-03-17
Remove Atmos Bin

plribeiro3000:
This file should not be included in `Fog` since it is already in the provider gem.


geemus:
Thanks!

PR #3491, opened 2015-03-18
[CloudSigma] add snapshots

zephyrean:
Added snapshots to compute model for CloudSigma provider. Snapshots are point-in-time versions of a drive. They can be cloned (promoted) to a full drive. [Official API docs] (https://cloudsigma-docs.readthedocs.org/en/latest/snapshots.html).

### DIFF
--- a/lib/fog/cloudsigma/models/snapshot.rb
+++ b/lib/fog/cloudsigma/models/snapshot.rb
@@ -1,0 +1,76 @@
+require 'fog/cloudsigma/nested_model'
+require 'fog/cloudsigma/models/volume'
+
+module Fog
+  module Compute
+    class CloudSigma
+      class Snapshot < Fog::CloudSigma::CloudsigmaModel
+        identity :uuid
+
+        attribute :timestamp, :type => :string
+        attribute :tags
+        attribute :status, :type => :string
+        attribute :resource_uri, :type => :string
+        attribute :permissions, :type => :array
+        attribute :owner
+        attribute :name, :type => :string
+        attribute :meta
+        attribute :grantees, :type => :array
+        attribute :drive
+        attribute :allocated_size, :type => :integer
+        
+        def save
+          if persisted?
+            update
+          else
+            create
+          end
+        end
+
+        def create
+          requires :name, :drive
+          data = attributes
+
+          response = service.create_snapshot(data)
+          new_attributes = response.body['objects'].first
+          merge_attributes(new_attributes)
+        end
+
+        def update
+          requires :identity, :name
+
+          data = attributes()
+
+          response = service.update_snapshot(identity, data)
+          new_attributes = response.body
+          merge_attributes(new_attributes)
+        end
+
+        def destroy
+          requires :identity
+
+          service.delete_snapshot(identity)
+
+          true
+        end
+
+        alias_method :delete, :destroy
+
+        def clone(clone_params={})
+          requires :identity
+          response = service.clone_snapshot(identity, clone_params)
+
+          Volume.new(response.body)
+        end
+        
+        alias_method :promote, :clone
+        
+        
+        def available?
+          status == 'available'
+        end
+        
+      end
+    end
+  end
+end
--- a/lib/fog/cloudsigma/models/snapshot.rb
+++ b/lib/fog/cloudsigma/models/snapshot.rb
@@ -18,7 +18,7 @@ module Fog
         attribute :grantees, :type => :array
         attribute :drive
         attribute :allocated_size, :type => :integer
-        
+
         def save
           if persisted?
             update
@@ -39,7 +39,7 @@ module Fog
         def update
           requires :identity, :name
 
-          data = attributes()
+          data = attributes
 
           response = service.update_snapshot(identity, data)
           new_attributes = response.body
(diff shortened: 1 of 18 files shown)


geemus:
@zephyrean - this is looking great so far. Could you look in to fleshing out tests for this a bit more? I'm concerned that otherwise it may be all too easy for us to introduce regressions or break things moving forward. Thanks!

zephyrean:
Thank you for your comment. I have updated this pull request with one more commit with tests for snapshots.

lib/fog/cloudsigma/models/snapshot.rb
```diff
@@ -0,0 +1,76 @@
...
+
+          response = service.create_snapshot(data)
+          new_attributes = response.body['objects'].first
+          merge_attributes(new_attributes)
+        end
+
+        def update
+          requires :identity, :name
+
+          data = attributes()
```
plribeiro3000: Could you remove this extra `()`?

lib/fog/cloudsigma/models/snapshot.rb
```diff
@@ -0,0 +1,76 @@
...
+          new_attributes = response.body
+          merge_attributes(new_attributes)
+        end
+
+        def destroy
+          requires :identity
+
+          service.delete_snapshot(identity)
+
+          true
```
plribeiro3000: What if it fails to delete the snapshot?

lib/fog/cloudsigma/requests/stop_server.rb
```diff
@@ -2,23 +2,24 @@ module Fog
   module Compute
     class CloudSigma
       class Real
-        def stop_server(server_id)
+        def stop_server(server_id, do_graceful_acpi = false)
```
zephyrean: Ok, will change it.

lib/fog/cloudsigma/models/snapshot.rb
zephyrean: What are your suggestions? I really don't know: all other providers in this library seem to do exactly like that.

plribeiro3000: I agree. The other providers do the exact same thing thus why i'm trying to push to improve it a little bit when possible. Do you think you can handle the request response code and then return `true` or `false` here based on that? Or CloudSigma does not return anything in case of error?

zephyrean: Their API returns a 404 which is then translated into exception by the library:
Fog::CloudSigma::Errors::NotFound: Object with uuid xxx does not exist
So this method either succeeds and returns 'true' or fails with exception

plribeiro3000: Well. Maybe we can rescue and return false then? What do you think?

lib/fog/cloudsigma/models/snapshot.rb
```diff
@@ -0,0 +1,76 @@
...
+        attribute :tags
+        attribute :status, :type => :string
+        attribute :resource_uri, :type => :string
+        attribute :permissions, :type => :array
+        attribute :owner
+        attribute :name, :type => :string
+        attribute :meta
+        attribute :grantees, :type => :array
+        attribute :drive
+        attribute :allocated_size, :type => :integer
```
plribeiro3000: I know this is just a detail but could you also do this in alphabetical order?
Once this list gets big we tend to repeat attributes, like we have in other places.

lib/fog/cloudsigma/models/snapshot.rb
zephyrean: But I am afraid that way user will not be able to know what caused this 'false' reply. It could be a simple 404 (meaning he was trying to delete a snapshot that did not exist) or it could be 5xx error message signalling of API server failure. These exceptions are used in this library just for that: they convey the reason of failure with type and meaningful error message as with the example I posted in the previous message.

plribeiro3000: Fair enough. 👍

lib/fog/cloudsigma/models/snapshot.rb
zephyrean: Ok, sure. Should I do the same in tests/cloudsigma/requests/*_tests.rb where we specify reply format?

plribeiro3000: Yes. That would be awesome!

zephyrean: I have pushed these changes

plribeiro3000: Awesome!

plribeiro3000:
LGTM. @zephyrean Could you squash those commits together so i can merge this?

thanks!

zephyrean:
Hi,
GitHub has recently added a new feature that allows you to squash commits upon merging:
https://help.github.com/articles/about-pull-request-merge-squashing
Will that help or should I squash commits manually?

plribeiro3000:
Lol. I always forget about it. Thanks for the reminder. =)

plribeiro3000:
Grat work here. Thanks for bearing with me on it!

zephyrean:
Thank you for your prompt comments and for maintaining this great library.

plribeiro3000:
👍

geemus:
Thanks!

geemus:
Thanks!

PR #3496, opened 2015-03-25
[glesys] fix URL encoding of API parameter values

tnn2:
Despite the name, the urlencode function does not actually apply URL encoding. This change fixes that.


geemus:
Good catch, thanks!

PR #3498, opened 2015-03-25
Adds RESTlike methods to Plan of Planning service

petrblaho:
Adds destroy, save, create and update methods to Plan
model of OpenStack Planning service.


geemus:
Thanks!

PR #3500, opened 2015-03-26
[GleSYS] add support for SSH key management

tnn2:
Implements support for SSH key management in the GleSYS module as per:
https://github.com/GleSYS/API/wiki/Full-API-Documentation#sshkeyadd

### DIFF
--- a/lib/fog/glesys/compute.rb
+++ b/lib/fog/glesys/compute.rb
@@ -14,6 +14,8 @@ module Fog
       model       :template
       collection  :ips
       model       :ip
+      collection  :ssh_keys
+      model       :ssh_key
 
       request_path 'fog/glesys/requests/compute'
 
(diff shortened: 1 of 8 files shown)


geemus:
I think in other parts of fog we have written this as `ssh_key` and `ssh_keys` (with an underscore in the middle). So for consistency I think maybe tweaking that might be a good call, what do you think?

tnn2:
The reason things are named this way is to be consistent with the upstream API, which names it's parameter "sshkey" for the add request but stores it in the field "data" in the model. I don't object to changing it to be consistent with fog nomenclature though, if you think that's more appropriate.

geemus:
Thanks for the clarification. I guess I would split it into two categories (this is what we do elsewhere in fog). For requests, keep the upstream spelling, these should map closely to one another. For collections/models though, we tend to try and follow the fog patterns (rather than upstream), so for those, the _ spelling would probably be preferred. This helps with interoperability for end users and keeps things consistent. Does that make sense? What do you think?

tnn2:
That sounds reasonable, I will update the branch over the weekend.

geemus:
Great, thanks!

tnn2:
Hi, I renamed some things, please review again. Thanks!

geemus:
Thanks!

PR #3501, opened 2015-03-28
Host and port options support for create_temp_url

biomancer:
I reused working code for building URL from `lib/fog/openstack/requests/storage/get_object_https_url.rb` in `lib/fog/hp/storage.rb` to add support for port there and to keep identical interface for both `create_temp_url` methods. I have not tested hp storage, but the fixes are trivial.

### DIFF
--- a/lib/fog/hp/storage.rb
+++ b/lib/fog/hp/storage.rb
@@ -300,6 +300,15 @@ module Fog
           @hp_secret_key = options[:hp_secret_key]
           @hp_tenant_id = options[:hp_tenant_id]
           @os_account_meta_temp_url_key = options[:os_account_meta_temp_url_key]
+
+          @hp_storage_uri = options[:hp_auth_uri]
+
+          uri = URI.parse(@hp_storage_uri)
+          @host   = uri.host
+          @path   = uri.path
+          @persistent = options[:persistent] || false
+          @port   = uri.port
+          @scheme = uri.scheme
         end
 
         def data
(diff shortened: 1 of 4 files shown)


geemus:
@biomancer looks like this breaks the CI build, could you take a look? Thanks!

biomancer:
@geemus Looks like the cause of test failure is that fog connection is not initialised properly before the test, there is no host set (and no scheme too if `create_temp_url` is called directly, not from `get_object_http_url` or `get_object_https_url`).

For example in tests for get_object_http_url the old result that was tested looked like that:
`http:///fogobjecttests/fog_object?temp_url_sig=205e6b18e05edd7b9193eeac02e87e72c1339cb9&temp_url_expires=1427737202`,
the new one is 
`http:/fogobjecttests/fog_object?temp_url_sig=205e6b18e05edd7b9193eeac02e87e72c1339cb9&temp_url_expires=1427737202`

None of them are valid, but the old one was passing all tests, and the new one fails on `start_with? "http://"` check. 

I'm not sure what I should do here - probably update test somehow so that the mocked connection includes host? And maybe raising an error in method if no host or scheme is set? Just forcing the old semi-correct behavior does not seem right.

Also there is a second problem - ruby 1.8.7 build fails as there is no method `URI::encode_www_form` for this version of stdlib, but this method is already used in master branch in `create_temp_url` for OpenStack, the tests for it are marked as pending, so they are missing it.

geemus:
Thanks for the details. Yeah, I think updating the tests to be more accurate sounds good. I guess for 1.8.7 we might need to copy/decompose that so we can use it directly?

biomancer:
@geemus I've updated the tests and and added some minor improvements, could you take a look please? The travis build is failing, but it seems that it's some upstream issue.

geemus:
Thanks!

PR #3506, opened 2015-04-01
Adding V3 and Domain openstack/core.rb and openstack/compute.rb

mikezuniga:
in order to be able to request tokens using V3 API and also to compute.rb in order to support requests for compute using V3 keystone style auth

### DIFF
--- a/lib/fog/openstack/core.rb
+++ b/lib/fog/openstack/core.rb
@@ -57,6 +57,10 @@ module Fog
       case options[:openstack_auth_uri].path
       when /v1(\.\d+)?/
         authenticate_v1(options, connection_options)
+      when /v2(\.\d+)?/
+        authenticate_v2(options, connection_options)
+      when /v3(\.\d+)?/
+        authenticate_v3(options, connection_options)
       else
         authenticate_v2(options, connection_options)
       end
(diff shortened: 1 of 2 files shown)


lib/fog/openstack/core.rb
```diff
@@ -169,13 +173,113 @@ def self.authenticate_v2(options, connection_options = {})
...
+    # Keystone Style Auth
+    def self.authenticate_v3(options, connection_options = {})
+      uri                   = options[:openstack_auth_uri]
+      tenant_name           = options[:openstack_tenant]
+      service_type          = options[:openstack_service_type]
+      service_name          = options[:openstack_service_name]
+      identity_service_type = options[:openstack_identity_service_type]
+      endpoint_type         = (options[:openstack_endpoint_type] || 'public').to_s
+      openstack_region      = options[:openstack_region]
+      domain_name           = options[:openstack_domain]
```
TerryHowe: Probably going to need user and project domain as well at least at some point.

geemus:
Looks like this got closed. Anything that I can help with?

mikezuniga:
I re submitted pull request its the latest one in the list

geemus:
Cool, I try to do FIFO, so just hadn't gotten that far yet, hopefully will get there pretty shortly.

clarsonneur:
thank you very much for this fix! i'm ok with core/compute update. that's what I need as well, at least, for now.

I think we may need to support user/project domain, as well. but not for now. 

thank you

lib/fog/openstack/core.rb
clarsonneur: I have tested this code, but it is failing in core.rb, as user domain and project domain should not be necessarily the same. (line 340, 350)

This certainly means that I need to split project and user domain. 

I can do the update and propose a merge request if you want.

geemus: Sounds good, thanks!

mikezuniga: Have you tried using it like this????

Fog::Compute.new({:provider => 'openstack', :openstack_tenant => 'tenant_id_here', :openstack_api_key => 'tenant_password', :openstack_username => 'myusername', :openstack_auth_url => 'https://keystone/v3/auth/tokens', :openstack_domain => 'Your_domain'})

retrieve_tokens_v3 takes the openstack_username, openstack_tenant and openstack_domain and gets you the token, then authenticate_v3 makes sure that your user is inside of the token by using that information from there.

clarsonneur: Yes, the problem is that the user and the project are not on the same domain.

if you set project scope with the user domain, it fails.

clarsonneur: My proposal user/domain split code for reference: https://github.com/fog/fog/pull/3518

PR #3514, opened 2015-04-03
Do not Merge yet! Remove google

plribeiro3000:
Once fog/fog-google#5 is merged and the gem is published this will be g2g.

cc/ @icco @erjohnso @ihmccreery


ihmccreery:
Okay, I've released a fog-google [gem](https://rubygems.org/gems/fog-google), and I've done some spot testing.  It looks like if I install a `gem build` of your branch, then I get the following behavior:

```
> require 'fog'
=> true
> connection = Fog::Compute.new({ provider: 'google', ... })
=> # huge blob of stuff
> server = connection.servers.bootstrap
=> # looks good
> server.destroy
> # also looks good
```

I'm willing to call this good.  However, I also have

```
> require 'fog-google'
LoadError: cannot load such file -- fog-google
```

Is this desired behavior for now?  I think eventually we'd like these provider gems to stand on their own, but I see similar behavior with other `fog-provider` gems.

ihmccreery:
Wait.  When I do a `bundle install` in the `fog` repository, it's installing `fog-google 0.0.1`, which doesn't have the binary.  Any ideas?

plribeiro3000:
You should be doing `fog/google`, not `fog-google`.

I will force the suage og version `0.0.2` to fix this issue. Give me a minute. =)

ihmccreery:
Lol.  New around here.

Version forcing actually isn't necessary, since no one should see 0.0.1 except me, since I yanked it, but maybe good anyway.  (I fixed my issue with `bundle update`.)

plribeiro3000:
Good to know its fixed for you then.

Beware `gem yank`. It is as bad as it seems. :skull:

ihmccreery:
Note taken.

ihmccreery:
So we're good to go on this, yes?

plribeiro3000:
Yep. The build just passed. =)

You are good to go on with `fog-google` now!

ihmccreery:
:herb: :tulip: :sunflower: :herb:

geemus:
Thanks!

PR #3517, opened 2015-04-04
[OpenStack|Network] Preliminary Meter Support

Matt1360:
This just goes over four of the [Neutron Meter](https://wiki.openstack.org/wiki/Neutron/Metering/Bandwidth) commands.  I'm using it for some stuff, and wanted to see if it would be useful to anyone else.

### DIFF
--- a/lib/fog/openstack/network.rb
+++ b/lib/fog/openstack/network.rb
@@ -131,6 +131,12 @@ module Fog
       request :update_quota
       request :delete_quota
 
+      # Meters
+      request :meter_label_create
+      request :meter_label_delete
+      request :meter_label_rule_create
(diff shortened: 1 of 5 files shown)


geemus:
Seems like a good start, but could use some tests around these?

Matt1360:
Yep! :)

I want to clean this up a little more, and maybe add a few more requests to it.  I'll try to add some tests in this week at the same time.

geemus:
Sounds great, just let me know if you have questions or need help. Thanks!

plribeiro3000:
@Matt1360 Could you open this against `fog-openstack` instead please?

Thanks!

PR #3519, opened 2015-04-07
[Brightbox] Remove provider specific helper test

tokengeek:
Before `fog-brightbox` was extracted, this tested the helper we use to
select the best image ID to use when testing.

However this is not `fog` specific anymore and so can be removed.

The remaining tests continue to exercise the main `fog` (abstracted)
interface and should remain for backwards compatibility.


tokengeek:
Errors appear to have been introduced for 1.8.7 by the HP provider and edge errors from the OpenStack "Baremetal" changes.

The baremetal change appears to have been caused by the require being missed out during a change in `fog-core`

geemus:
Thanks!

On Tue, Apr 7, 2015 at 10:35 AM, Paul Thornthwaite <notifications@github.com

PR #3520, opened 2015-04-07
[openstack|network,compute,core] Supporting openstack auth v3 - project/user domain

clarsonneur:
* Splitting :openstack_domain to :openstack_userdomain and :openstack_prjdomain
* Added :openstack_tenantid which can replace authentication scope from :openstack_prjdomain + openstack_tenant to :openstack_tenantid.

### DIFF
--- a/lib/fog/openstack/compute.rb
+++ b/lib/fog/openstack/compute.rb
@@ -6,10 +6,11 @@ module Fog
       requires :openstack_auth_url
       recognizes :openstack_auth_token, :openstack_management_url,
                  :persistent, :openstack_service_type, :openstack_service_name,
-                 :openstack_tenant,
+                 :openstack_tenant, :openstack_tenant_id,
                  :openstack_api_key, :openstack_username, :openstack_identity_endpoint,
                  :current_user, :current_tenant, :openstack_region,
-                 :openstack_endpoint_type
+                 :openstack_endpoint_type,
+                 :openstack_prj_domain, :openstack_user_domain
 
       ## MODELS
       #
@@ -250,10 +251,12 @@ module Fog
 
         def initialize(options={})
           @openstack_username = options[:openstack_username]
-          @openstack_domain = options[:openstack_domain]
+          @openstack_user_domain = options[:openstack_user_domain] || options[:openstack_domain]
+          @openstack_prj_domain  = options[:openstack_prj_domain]  || options[:openstack_domain] || 'Default'
           @openstack_auth_uri = URI.parse(options[:openstack_auth_url])
 
           @current_tenant = options[:openstack_tenant]
+          @current_tenant_id = options[:openstack_tenant_id]
 
           @auth_token = Fog::Mock.random_base64(64)
           @auth_token_expiration = (Time.now.utc + 86400).iso8601
@@ -290,7 +293,9 @@ module Fog
         attr_reader :auth_token_expiration
         attr_reader :current_user
         attr_reader :current_tenant
-        attr_reader :openstack_domain
+        attr_reader :current_tenant_id
+        attr_reader :openstack_user_domain
+        attr_reader :openstack_prj_domain
 
         def initialize(options={})
           @openstack_auth_token = options[:openstack_auth_token]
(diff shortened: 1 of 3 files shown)


mdarby:
Tests?

clarsonneur:
build failed, I did no change on the area impacted in the build failure. So, I would need help to understand what is wrong.

At least, the test done on my side is good.

tokengeek:
@clarsonneur - The CI issue has probably been fixed with https://github.com/fog/fog-core/commit/aa02864717c2bf1d50a3e0b9074ed04a9257d354

The baremetal service was missed so the edge version of `fog-core` was broken briefly.

clarsonneur:
@tokengeek ok cool! So, what should I do? wait?

tokengeek:
I've kicked off the travis builds, hopefully that will clear up the errors. Then other folks can review the PR

clarsonneur:
Yes it did. great! build successful! thank you for you help. waiting for other to review then.

geemus:
@tokengeek thanks!

geemus:
I wonder a little about underscoring, ie

* `openstack_tenantid` vs `openstack_tenant_id`
* `openstack_prjdomain` vs `openstack_prj_domain`
* `openstack_userdomain` vs `openstack_user_domain`

Also, I don't know what prj is, but maybe that just reflects my relatively fleeting knowledge of openstack.

@mdarby could you take another look?

clarsonneur:
I can add extra _ no problem for that.

clarsonneur:
About Project domain, there is several discussions about that:
One example: http://www.madorn.com/keystone-v3-api.html#.VSTrgctTl6U
The openstack auth v3 API has also a lot of things about domain, now.
{url}//v3/domains
http://developer.openstack.org/api-ref-identity-v3.html

clarsonneur:
Added extra _

geemus:
@clarsonneur thanks, seems a bit nitpicky even to me, but the consistency slips add up if you don't keep an eye on them.

geemus:
Thanks!

clarsonneur:
agree. Thank you! 

Usually, when are you creating a new gem version with those latest merge? I saw it to be something like one new version per month. So, may be, next month should be the plan? as 1.29 was published one week ago...

geemus:
Yeah, to make it more expectable, barring major bugs or whatever we are shooting for the first thursday of each month.

lib/fog/openstack/compute.rb
```diff
@@ -6,10 +6,11 @@ class OpenStack < Fog::Service
       requires :openstack_auth_url
       recognizes :openstack_auth_token, :openstack_management_url,
                  :persistent, :openstack_service_type, :openstack_service_name,
-                 :openstack_tenant,
+                 :openstack_tenant, :openstack_tenant_id,
                  :openstack_api_key, :openstack_username, :openstack_identity_endpoint,
                  :current_user, :current_tenant, :openstack_region,
-                 :openstack_endpoint_type
+                 :openstack_endpoint_type,
+                 :openstack_prj_domain, :openstack_user_domain
```
TerryHowe: While I don't like excessive typing, what I don't like more is odd ball abbreviations I cannot remember.  Can we get openstack_project_domain?

lib/fog/openstack/network.rb
```diff
@@ -310,18 +323,29 @@ def request(params)
...
             }
 
-            credentials = Fog::OpenStack.authenticate_v2(options, @connection_options)
+            case @openstack_auth_uri.path
+            when %r{/v2.0}
+              credentials = Fog::OpenStack.authenticate_v2(options, @connection_options)
+            when %r{/v3}
+              credentials = Fog::OpenStack.authenticate_v3(options, @connection_options)
+            else
+              credentials = Fog::OpenStack.authenticate_v2(options, @connection_options)
```
TerryHowe: I'd be tempted to make v3 auth default

lib/fog/openstack/compute.rb
clarsonneur: I understand. I can make a small commit with that fix

lib/fog/openstack/network.rb
clarsonneur: yeah. I'm not sure which one to select either if the path do not give the information. the thing is v3 means require a domain, while v2 do not expect this.  less constraint was my decision. but I may be wrong.

PR #3521, opened 2015-04-08
Extract fog-libvirt

strzibny:
- adds `fog-libvirt` as a new dependency in gemspec
- removes `libvirt` provider specific files


strzibny:
This will pass on Travis once we release `fog-libvirt` gem. (Checked locally.)

geemus:
@strzibny cool! Have you setup fog-libvirt somewhere? Would you be ok with us moving it under the org? If you give me admin I can do that. Thanks!

strzibny:
@geemus not sure if collaborators would do, do I need to 'Transfer ownership'? Don't know how to make you admin without that.

geemus:
@strzibny I do it infrequently enough to never remember the specifics. It doesn't look like I have permissions enough as a collaborator. Maybe just transfer it to me and then I'll transfer it in to the org? (you wouldn't be able to just transfer to the org, because you are not an admin of the org, if I remember right anyway). Permissions are the worst!

geemus:
I think it should be cut over and you should have an invite to the admin team for fog-libvirt team.

strzibny:
Yes, I did. Thanks.

geemus:
I guess now we should just see about fog-libvirt released and then we should be ready to bring this in. If you do that, please also add myself and @tokengeek as owners on the gem (so we can push if/when needed). Thanks!

strzibny:
Alright, I do that.

geemus:
Thanks!

strzibny:
Done. You should both be admins. I will do a Fedora package now and look into replacing the dep in `vagrant-libvirt` and check that it works as expected.

plribeiro3000:
I think we can close this then. Thank you!

strzibny:
@plribeiro3000 So, you still want to keep the files around in `fog` as well? Why?

plribeiro3000:
Damm. Wrong push up. I thought this was just an issue to do the job. sorry about that.

plribeiro3000:
@strzibny It seems it is missing the libvirt package in the `travis` server. Could you take a look at it?

I think that you might have to tell `travis` to install the lib in the [`.travis.yml`](https://github.com/fog/fog/blob/master/.travis.yml) file.

strzibny:
@plribeiro3000 seemed like that fog-libvirt has not been enabled to trigger builds on travis, so I enabled it (next push should trigger the build).

plribeiro3000:
@strzibny I was talking about the build in this PR. It is not passing because it does not find the `libvirt` library in the travis machine running the tests.

We need to make sure `fog` behavior will be the same after we extract the provider.

strzibny:
I see, fixed. Now the build failes because I removed https://github.com/strzibny/fog/commit/ec1162682814e460bf419fb6efe0afd6807122ae. I wasn't sure about this file, I guess I should include it back?

plribeiro3000:
I think that perhaps you should wrap the dependency of `fog-libvirt` in the `fog.gemspec` like this:

```ruby
if ENV["FOG_USE_LIBVIRT"]
    s.add_dependency("fog-libvirt")
end
```

and remove the dependency on `ruby-libvirt` as well.

strzibny:
I understand that devel dep on `ruby-libvirt` can be removed, but are you sure that fog as a gem should depend on fog-libvirt only if an ENV var is present? That would mean that you could build fog gems with and without libvirt support.

geemus:
@strzibny good question. Maybe actually this indicates that we should NOT have a dependency on `fog-libvirt` by default (since it would imply the ``ruby-libvirt` dep as well). Then maybe you could just do builds on fog-libvirt and not worry as much about fog proper? What do you think?

plribeiro3000:
@strzibny @geemus Agreed. `fog-libvirt` seems to be an exception to our current build.

strzibny:
@geemus @plribeiro3000 does that mean getting rid of all related test files/code as well?

geemus:
@strzibny um, yeah I think so? Ideally that stuff would all just live in the new repo if I understand correctly. Does that seem reasonable/make sense?

strzibny:
Yes. In this PR I already removed everything. I have one question about `lib/fog/bin/libvirt.rb
` file (https://github.com/strzibny/fog/commit/ec1162682814e460bf419fb6efe0afd6807122ae). How this stuff work? Is it necessary for `fog-libvirt`? I have seen other extracted fog gems and they did not take the bin file with them.

geemus:
Great question. It depends on usage patterns I think. It is just meant as a convenience/helper, so probably not needed (if I remember correctly).

@plribeiro3000 did you have any further thoughts/concerns there?

geemus:
@strzibny thanks again for helping on this, just let us know if you have further questions or anything we can help with. Good to see this getting sorted out (or at least well on it's way).

plribeiro3000:
It is necessary @strzibny. All extracted providers should have it as we are moving provider specific logic out of `fog`.
Which other extracted providers don't have a bin file?

strzibny:
Sorry, I got confused I guess. I have that file in `fog-libvirt`, so everything should be alright. Thanks guys.

geemus:
Thanks!

Issue #2954, opened 2014-05-26
[vsphere] modify vm_clone request to perform a windows sysprep on the image


rob12:
I need to modify the Vsphere vm_clone request so that when a windows VM is cloned, it is sysprep'd, and joined to an AD domain.  

In summary, this will involve replacing the current RbVmomi::VIM::CustomizationLinuxPrep object that is passed to the Vsphere API with an RbVmomi::VIM::CustomizationSysprep (parameter based), or RbVmomi::VIM::CustomizationSysprepText (to use an unattend.xml rendered from ERB).

Would the maintainers prefer that this work be done in the current vm_clone request (making decisions on the OS type), or would it be better to create something like a vm_win_clone request?

Any other requirements?

sofam:
@rob12 also looking for info on how to pass the customization spec to foreman!

fherbert:
I've modified to vm_clone.rb and cloudinit_to_customspec.rb to support the full vmware customization spec. I've submitted a pull request https://github.com/fog/fog/pull/3523

geemus:
Awesome, thanks!

plribeiro3000:
It seems that this issue was resolved with a PR to support full vmware customization spec at #3253

Closing.

Comment back if this is not the case.

Thanks!

PR #3523, opened 2015-04-09
Support the full vmware customization spec #2954

fherbert:
Add support for the full vmware customization spec as per https://pubs.vmware.com/vsphere-55/index.jsp#com.vmware.wssdk.apiref.doc/vim.vm.customization.Specification.html


fherbert:
@nirvdrum Anything I can do to help get this merged into master?

nirvdrum:
Sorry.  I'll try to take a look at this soon.  Life was conspiring against me.  I got through most of the rest of the vSphere backlog and hope to be fully clear this week.

nirvdrum:
Having briefly looked though, do you know whether these changes are backwards-compatible?  There are fog users going back to at least vSphere 4.  And there are a lot on 5.1.

fherbert:
Looks like the customization spec has been around for some time, the earliest reference I can find for it is in SDK Package for VMware VirtualCenter 1.2. It's present in the Vsphere 4 SDK (https://www.vmware.com/support/developer/vc-sdk/visdk400pubs/ReferenceGuide/index.html)

nirvdrum:
Yeah, the spec has been around, but the fields might have changed.  I just wanted to make sure that that didn't change in an incompatible manner.  If not, fantastic.

fherbert:
I've just done a quick check, all the same options are there in the 4.0 SDK.

nirvdrum:
That's very helpful, thanks.  I'm on my laptop without access to my vSphere cluster, so I can't check currently.  But I'll give this a shot this week.  I give you permission to spam me if I don't :-)

fherbert:
I've done a quick write up on how to use this (from Foreman) but there's
some sample yaml to use as a template for the customization spec here
http://www.herbert.org.nz/?p=245
There is one niggle I'm not 100% sure about is if the template has an
existing nic and you don't pass the nicsetting map, vshpere will complain.
So you can see in the yaml I send some dummy data so the nicsetting map
dataype gets created.

nirvdrum:
I didn't notice any regressions in my own testing.  Unfortunately, it's really hard to say how this might impact others.  But Foreman is a big consumer of fog's vsphere support, so having it drive the work here helps assuage concerns a lot.

nirvdrum:
I don't know if this is something you can apply to foreman, but I've found it easiest to clone the VM, delete the NICs, create new NICs, and then run the customize task against the new NICs.  It avoids any headaches associated with cardinality of the template's NICs.  And it helps ensure the NIC you get is of the type you want on the network you want.  You can see how it's managed in Rubber here:

https://github.com/rubber/rubber/blob/e5ff6b56374038cbefedf82fd8468a12d2ddd7c7/lib/rubber/cloud/vsphere.rb#L81-L96

geemus:
Thanks!

PR #3525, opened 2015-04-09
Hash#reject! returns nil if no changes were made; it is not correct beha...

djaara:
`Hash#reject!` returns `nil` if no changes were made; it is not correct behaviour for config assignment

If there is not any `nil` value in `Hash`, nil is returned instead of expected configuration and thus it subsequent API call fails.

### DIFF
--- a/lib/fog/xenserver/requests/compute/create_server.rb
+++ b/lib/fog/xenserver/requests/compute/create_server.rb
@@ -85,7 +85,7 @@ module Fog
 
         def create_server( name_label, template = nil, networks = [], extra_args = {})
           if name_label.is_a?(Hash)
-            config = name_label.reject! { |_k,v| v.nil? }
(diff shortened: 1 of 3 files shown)


geemus:
Thanks!

plribeiro3000:
@djaara Good catch. I think i've already solved that issue in [`fog-xenserver`](https://github.com/fog/fog-xenserver/blob/master/lib/fog/compute/xen_server/models/instance_methods.rb#L24) without knowing of it.

BTW, take a look at `fog-xenserver` since the code here is going to be removed and `fog` will depend on `fog-xenserver` soon which has some breakings regarding the api.

djaara:
@plribeiro3000 :+1:  I found [fog-xenserver](https://github.com/fog/fog-xenserver) when I looked around for other pull requests, I'll give it a try.

PR #3529, opened 2015-04-14
[XenServer] Fix the 'redirect to master' functionality

kongslund:
When authenticating towards a slave host, the error `HOST_IS_SLAVE` is returned by the slave host together with the IP address of the master host. Previously this resulted in an `InvalidLogin` exception which was hiding the actual cause. Now it results in an `HostIsSlave` exception which is then used to redirect the connection to the master host in case `:xenserver_redirect_to_master` is true.

This fix also removes the previous behaviour of doing a double login when the host is a master.

This is a breaking change since I have removed the `find_pool_master` method from the `Fog::XenServer::Connection` class. However it was not working (it invoked `authenticate` at the beginning which resulted in an `InvalidLogin` exception in case of a slave) and the official way is to let `Fog::Compute::XenServer` deal with the connection.

### DIFF
--- a/lib/fog/xenserver/compute.rb
+++ b/lib/fog/xenserver/compute.rb
@@ -126,17 +126,18 @@ module Fog
           @timeout              = options[:xenserver_timeout] || 30
           @redirect_to_master   = options[:xenserver_redirect_to_master] || false
           @connection           = Fog::XenServer::Connection.new(@host, @timeout)
-
-          if @redirect_to_master == false
-            @connection  = Fog::XenServer::Connection.new(@host, @timeout)
-          elsif @redirect_to_master == true
-            host_master  = @connection.find_pool_master(@username, @password)
-            if host_master && host_master!= @host
-              @host = host_master
+          begin
+            @connection.authenticate(@username, @password)
+          rescue Fog::XenServer::HostIsSlave => e
+            if @redirect_to_master
+              Fog::Logger.debug "Redirecting to master #{e.master_ip}"
+              @host = e.master_ip
               @connection  = Fog::XenServer::Connection.new(@host, @timeout)
+              @connection.authenticate(@username, @password)
+            else
+              raise e
             end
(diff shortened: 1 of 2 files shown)


kongslund:
@plribeiro3000: I've noticed that [fog-xenserver](https://github.com/fog/fog-xenserver) [lacks](https://github.com/fog/fog-xenserver/blob/master/lib/fog/compute/xen_server/real.rb) the 'redirect to master' functionality. Wouldn't it be nice to have as well?

plribeiro3000:
@kongslund Hard to know. I'm not sure this is something that happens so often that needs to be pushed towards the lib itself instead of the user code.
If we need/want this feature i believe it should be in [`lib/fog/xen_server/connection`](https://github.com/fog/fog-xenserver/blob/master/lib/fog/xen_server/connection.rb#L14-L18). 
Do you have any thoughts about it?

Also, i'm open to listen to another ideas that you might have.

kongslund:
@plribeiro3000 The changes to `lib/fog/xenserver/core.rb` could be ported no matter what. The error returned is more explicit - and more importantly - not misleading.

I don't know how many people are actually using the pooling capability of XenServer but whoever does should benefit from the redirect functionality since a master is not guaranteed to stay master forever.

Regarding putting everything in the `Connection` class: I believe that it would be a nice abstraction and much in line with what others have done in terms of failover capable database drivers. It could be a specialized `PoolConnection` class that could take a list of hosts and potentially also different credentials per host. Requests invoked through the connection are automatically sent to the master. If the master changes then it is transparent to the client of the connection. Likewise if new members are added to the pool. I do see some design/implementation challenges though. :-)

plribeiro3000:
@kongslund Agree.
Can you try to add at least a draft for this feature in `fog-xenserver` so we can start discussing on top of code?

Thanks!

geemus:
Thanks!

PR #3530, opened 2015-04-14
Added vsphere_debug flag for stderr debugging

lzap:
Unfortunately rbvmomi does not support environment variables, this is the only
way how to pass debug flag in to see request/reply on the STDERR.


lzap:
@shlomizadok opinion?

shlomizadok:
:+1: nice!

geemus:
Thanks!

PR #3535, opened 2015-04-19
[GleSYS] server/create: sync required and optional args with API spec

tnn2:
This syncs the list of required and optional arguments for the server/create command with the API specification:
https://github.com/GleSYS/API/wiki/Full-API-Documentation#servercreate

### DIFF
--- a/lib/fog/glesys/models/compute/server.rb
+++ b/lib/fog/glesys/models/compute/server.rb
@@ -74,13 +74,11 @@ module Fog
               :disksize       => disksize     || "10",
               :memorysize     => memorysize   || "512",
               :cpucores       => cpucores     || "1",
-              :rootpassword   => rootpassword,
-              :transfer       => transfer     || "500",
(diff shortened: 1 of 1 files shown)


plribeiro3000:
Hey @tnn2. I'm willing to bring this in. My only concern is that this is breaking the current behavior and users might be expecting this behavior so perhaps we should add a deprecation notice before removing `bandwidth` and `transfer` default values. What do you think?

tnn2:
Hi,
The `transfer` option has already been deprecated and ignored by the provider for at least a couple of months. The `bandwidth` option is defaulted by upstream to the same value as was previously set in fog (10). Based on this I don't believe it is likely to break anything.

geemus:
@tnn2 thanks for the clarification, bringing this in.

plribeiro3000:
@tnn2 Yeah. It makes sense. Thank you!

PR #3537, opened 2015-04-21
Change openstack_prj to openstack_project

TerryHowe:
It is hard to remember odd abbreviations.


TerryHowe:
@clarsonneur

geemus:
Seems good. Maybe we should support the old way also for backwards compatibility? (I can't recall now if the old way made it into a release, if not this is probably not a big deal and maybe isn't that big of a deal anyway). What do you think?

TerryHowe:
I don't think this v3 auth has been released yet, so no need for backwards compat.

geemus:
Sounds good. Thanks!

clarsonneur:
hello! thank you for this! Sorry, as I was out for sometimes. So I was not able to do it myself.

PR #3545, opened 2015-04-29
Fixes #3543, adds support for cpuHotAdd and memoryHotAdd

### LINKED ISSUE #3543: [vsphere] Support more options as per vsphere API
I'd like to modify vm_clone.rd and create_vm.rb to support more of the options available (https://www.vmware.com/support/developer/converter-sdk/conv43_apireference/vim.vm.ConfigInfo.html). The first ones I'd like to add is the ability to manage Memory and CPU hot add enable.
This will also involve changing lib/fog/vsphere/models/compute/server.rb and lib/fog/vsphere/compute.rb to enable to enable the retrieval of these values from vsphere.

fherbert:


geemus:
Thanks!

PR #3550, opened 2015-05-01
[vCloud Director] Fixing Unrecognized argument: path warning message

migmartri:
If you set the path option during instantiation, the console returns an"Unrecognized argument: path warning message". 

This commit removes that warning message.

selection_379


geemus:
Thanks!

PR #3551, opened 2015-05-01
[vCloud Director] Set oauth-token using an attr writer

migmartri:
This patch allows to set the vcloud_token as an instance attribute. This does not replace the current fog['FOG_VCLOUD_TOKEN'] approach just offers an alternative.

This makes easier to have multiple instances of Fog::Compute::VcloudDirector without mixing ENV variables.

Example of instantiation:

    @fog ||= Fog::Compute::VcloudDirector.new(
      vcloud_director_username:      vmware_api_username,
      vcloud_director_password:      vmware_password || "not-set",
      ....
    )

    @fog.vcloud_token = token

### DIFF
--- a/lib/fog/vcloud_director/compute.rb
+++ b/lib/fog/vcloud_director/compute.rb
@@ -44,7 +44,7 @@ module Fog
       class TaskError < Fog::VcloudDirector::Errors::TaskError; end
 
       requires :vcloud_director_username, :vcloud_director_password, :vcloud_director_host
-      recognizes :vcloud_director_api_version, :vcloud_director_show_progress, :path
+      recognizes :vcloud_director_api_version, :vcloud_director_show_progress, :path, :vcloud_token
 
       secrets :vcloud_director_password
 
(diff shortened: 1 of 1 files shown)


geemus:
@migmartri seems like a good thing to have. Instead of doing it this way, how would you feel about adding it to the `recognizes` list and having the instance var set in the initializer instead? I think that would be a bit more consistent with other fog providers. Then if you wanted to connect with a different token, you would instantiate a new object to use (instead of changing it like this). What do you think?

migmartri:
@geemus my first approach was based on the way we were using the library on our project.

It was easier for us to:

* instantiate fog
* call to a refresh_token(fog) if needed (not expired)
* Assign the new token to fog   (fog.vcloud_token = token)

Once done, consecutive calls will not refresh the token after calling the auth method.

I understand your concern, and I agree with you that it might be a better idea to do it in the initialize. The latest commit does exactly that.

Thanks for your feedback.

geemus:
Looks good I think, if you could rebase I think we should be able to get this in. Thanks!

geemus:
Thanks!

PR #3552, opened 2015-05-01
[vCloud Director] VM name and storage profile configuration on vApp instantiate.

migmartri:
Patch that adds the functionality of modifying the name and the storage profile of the VM's contained in the VApp template (catalog_item class) during instantiation.

As an example of use: 

        options = {
          vdc_id: VDC_ID, network_id: NETWORK_ID,
          # New section that allows to configure name and disk type
          vms_config: [
            {
              href: VM_SOURCE_HREF,
              storage_profile_href: STORAGE_PROFILE_HREF,
              name: "My VM name"
            }
          ]
        }

        catalog_item.instantiate(display_name, options)

If you do not add the vms_config the behavior is the same one than the current one. 

It is based on [this documentation](http://pubs.vmware.com/vca/index.jsp?topic=%2Fcom.vmware.vcloud.api.doc_56%2FGUID-BF9B790D-512E-4EA1-99E8-6826D4B8E6DC.html).
 
Regards,
Miguel

### DIFF
--- a/lib/fog/vcloud_director/requests/compute/instantiate_vapp_template.rb
+++ b/lib/fog/vcloud_director/requests/compute/instantiate_vapp_template.rb
@@ -61,6 +61,21 @@ module Fog
             }
             # The template
             xml.Source(:href => options[:template_uri])
+            # Use of sourceItems for configuring VM's during instantiation.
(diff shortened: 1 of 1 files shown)


geemus:
Thanks!

PR #3556, opened 2015-05-04
Fixes #3555, add firmware support for vsphere

### LINKED ISSUE #3555: [vsphere] add support for boot firmware
Similar to #3543, I would like to have support for changing the boot firmware when creating virtual machines.  Most (all?) non-Mac operating systems use BIOS by default with vSphere, but EFI is supported and may be preferred in some situations.

See 'firmware' at http://pubs.vmware.com/vsphere-60/index.jsp?topic=%2Fcom.vmware.wssdk.apiref.doc%2Fvim.vm.ConfigInfo.html, it has been supported since vSphere 5.0 and the two options are 'bios' and 'efi'

keeleysam:


geemus:
Thanks!

PR #3558, opened 2015-05-06
Allow passing query options to snapshots endpoint.

seanhandley:
This is so we can pass `all_tenants: true` to the endpoint and get a full list of snapshots for the platform.


TerryHowe:
lgtm

geemus:
Thanks!

PR #3561, opened 2015-05-12
[openstack|compute] added support for flavor extra_specs

tzumainn:
Adds support for getting and creating extra_specs for an OpenStack
flavor.

### DIFF
--- a/lib/fog/openstack/compute.rb
+++ b/lib/fog/openstack/compute.rb
@@ -98,6 +98,10 @@ module Fog
       request :create_flavor
       request :delete_flavor
 
+      # Flavor Actions
+      request :get_flavor_extra_specs
+      request :create_flavor_extra_specs
+
       # Flavor Access
       request :add_flavor_access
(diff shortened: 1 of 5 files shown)


mdarby:
:+1:

plribeiro3000:
hey @tzumainn. Could you add some tests around this? Thanks!

TerryHowe:
This might be one of those areas that Fog should hide the inconsistency of the REST API naming and call this flavor metadata or something.

tzumainn:
@plribeiro3000 Ah, yep, I'll add tests in
@TerryHowe That seems reasonable - should I go ahead and do that now, or do others need to weigh in as well?

tzumainn:
@plribeiro3000 tests added

plribeiro3000:
Seems pretty simple to me. I'm not that into Openstack but since @mdarby says its ok i will go ahead and merge!

Thank you @tzumainn !

geemus:
Thanks!

PR #3563, opened 2015-05-13
OpenStack rename flavor extra specs to metadata

TerryHowe:


TerryHowe:
@tzumainn the metadata rename thing I was thinking about.

TerryHowe:
The purpose of this is to hide the inconsistency of the REST API and call flavor extra specs metadata which is more consistent with other OpenStack resources.

mdarby:
Should :metadata be aliased :extra_specs for BWC?

TerryHowe:
@mdarby they extra spec PR was merged yesterday.  it hasn't been released.

mdarby:
Ah, I understand. Sorry for the noise.

geemus:
Thanks!

PR #3564, opened 2015-05-14
OpenStack allow domains to be optional fix warning

TerryHowe:
Allow project and user domains to be optional and make
openstack_project_name a valid option with no warning.


geemus:
Thanks!

PR #3566, opened 2015-05-18
[cloudstack] Allow specifying the size of DATADISK on the creation of a server

yyuu:
It seems that the `size` option is not passed to `deployVirtualMachine`. It is an option supported by CloudStack to specify the size of the data disk.

https://cloudstack.apache.org/docs/api/apidocs-4.5/user/deployVirtualMachine.html


plribeiro3000:
Thanks!

geemus:
Thanks!

PR #3567, opened 2015-05-22
OpenStack add nova service support

Ladas:
OpenStack add nova service API support.


Ladas:
Working on running the tests with real env

geemus:
Looks like a good start, just let me know how things are looking when you get a chance to work against real.

Ladas:
@geemus real tests are passing now for services

TerryHowe:
lgtm

geemus:
Thanks!

PR #3570, opened 2015-05-26
Vsphere - customizing interfaces and disks when cloning from template

iNecas:
After this patch, the clone_vm accepts interfaces and volumes options, taking
interface resp. volume objects to modify the ones defined in the template.
This allows to customize all the interfaces (not just changing the
network_label on one of them), as well as increasing the size of the volumes
and adding or removing them.

### DIFF
--- a/lib/fog/vsphere/requests/compute/list_vm_interfaces.rb
+++ b/lib/fog/vsphere/requests/compute/list_vm_interfaces.rb
@@ -28,8 +28,8 @@ module Fog
   #macAddress: "00:50:56:a9:00:28",
   #unitNumber: 7,
   #
-        def list_vm_interfaces(vm_id)
-          get_vm_ref(vm_id).config.hardware.device.grep(RbVmomi::VIM::VirtualEthernetCard).map do |nic|
+        def list_vm_interfaces(vm_id, datacenter = nil)
+          get_vm_ref(vm_id, datacenter).config.hardware.device.grep(RbVmomi::VIM::VirtualEthernetCard).map do |nic|
             {
               :name    => nic.deviceInfo.label,
               :mac     => nic.macAddress,
--- a/lib/fog/vsphere/requests/compute/vm_clone.rb
+++ b/lib/fog/vsphere/requests/compute/vm_clone.rb
@@ -68,6 +68,14 @@ module Fog
         #       Default true
         #     * 'time_zone'<~String> - *REQUIRED* Only valid linux options
         #       are valid - example: 'America/Denver'
+        #     * 'interfaces' <~Array> - interfaces object to apply to
+        #        the template when cloning: overrides the
+        #        network_label, network_adapter_device_key and nic_type attributes
+        #     * 'volumes' <~Array> - volumes object to apply to
+        #        the template when cloning: this allows to resize the
+        #        existing disks as well as add or remove them. The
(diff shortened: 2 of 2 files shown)


geemus:
@nirvdrum could you review? thanks!

nirvdrum:
I'm going to need to chew this one over.  While I agree the default behavior of cloning VMs is not ideal, that's just how the vSphere API works.  The expected workflow is you clone then customize the VM.  fog doesn't make the customization part any easier than what rbvmomi provides, so I appreciate the work you've done here.  But I'm wondering if it should be a separate set of requests rather than placing it in `vm_clone`, which is already an incredibly complex method.

Also, was the change to `list_vm_interfaces` necessary to get this working?  If not, I'd prefer to see that as a different commit, preferably even a different pull request.

iNecas:
I agree that the `vm_clone` is complex (I tried at least to extract the logic into separate methods, to have at least some structure there).

I would really like to do it as the one-request thing: since the VSphere is able to do so right away, I would like to avoid implementing the same logic on my side: with this approach, I don't need to wait for the cloning operation in my ruby process to finish, dealing with the things about not completing the changes, if my script gets killed etc, especially when the VSphere API provides this possibility.

The list_vm_interfaces change is there, because I use it form the vm_clone, and the datacenter is used for getting the vm_ref for the template https://github.com/fog/fog/pull/3570/files#diff-8d8a8319291f90a3f90d65a4e808b263R90: I wanted to stay consistent here.

nirvdrum:
Okay.  I misread this.  I thought you were cloning and then modifying the devices afterwards.

Let me pull this down locally and play with it a bit.  It'll probably be a few days before I can get to it.  But if it can be done in a single atomic operation, I agree that would be preferable.  I was under the impression the vSphere API didn't allow that, which appears to be incorrect.

iNecas:
@nirvdrum hi, any news on this one?

nirvdrum:
@iNecas Sorry, I haven't, and I probably won't until the end of this week.  My daughter was born yesterday, so it's been a bit crazy around here for the past week or so.

iNecas:
@nirvdrum no problem, congratulations!

geemus:
@nirvdrum congrats!

iNecas:
Ping

ohadlevy:
@nirvdrum :+1: from my side too.

nirvdrum:
Sorry.  Plowing through my backlog now.

nirvdrum:
And thanks for the warm wishes (and corresponding patience) :-)

ohadlevy:
@nirvdrum another ping perhaps? :)

nirvdrum:
My sincerest apologies for taking so long to get back to this.  I just pulled the branch and tried it out on my typical workflow and there were no regressions.  My automated steps do a clone then delete all NICs and recreate them, also creating new volumes.  So I didn't get to try out all the new customization options, but ensuring things still work the way they used to was the primary concern.  I trust the customization options work the way you guys expect :-)  I do look forward to trying them out, but I didn't want to hold this up any longer.

Thanks again!  And thanks for being so patient.

ohadlevy:
Thanks!

iNecas:
Thanks @nirvdrum !

geemus:
Thanks!

PR #3568, opened 2015-05-25
Openstack Ironic maintenance mode

Ladas:
Openstack Ironic maintenance mode, adding set/unset maintenance
actions.

PR #3571, opened 2015-05-27
[openstack|baremetal] added support for setting node states

tzumainn:
Adds support for setting node power and provision states

### DIFF
--- a/lib/fog/openstack/baremetal.rb
+++ b/lib/fog/openstack/baremetal.rb
@@ -34,6 +34,8 @@ module Fog
       request :list_nodes
       request :list_nodes_detailed
       request :patch_node
+      request :set_node_power_state
+      request :set_node_provision_state
 
       # Chassis requests
       request :create_chassis
@@ -69,8 +71,6 @@ module Fog
       # request :set_boot_device
       # request :list_supported_boot_devices
       # request :list_node_states
-      # request :set_power_state
-      # request :provision_node
       # request :get_console_info
       # request :change_console_state
       # request :get_vendor_passthru_methods
--- a/lib/fog/openstack/requests/baremetal/set_node_power_state.rb
+++ b/lib/fog/openstack/requests/baremetal/set_node_power_state.rb
@@ -1,0 +1,35 @@
+module Fog
+  module Baremetal
+    class OpenStack
+      class Real
+        def set_node_power_state(node_id, power_state)
+          data = {
+            'target' => power_state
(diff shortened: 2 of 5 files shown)


lib/fog/openstack/baremetal.rb
```diff
@@ -34,6 +34,8 @@ class OpenStack < Fog::Service
       request :list_nodes
       request :list_nodes_detailed
       request :patch_node
+      request :set_node_power_state
+      request :set_node_provision_state
```
TerryHowe: These should be set_power_state and set_provision_state and it would be best to clean up the comments:

      # request :set_power_state
      # request :provision_node

lib/fog/openstack/requests/baremetal/set_node_power_state.rb
```diff
@@ -0,0 +1,35 @@
+module Fog
+  module Baremetal
+    class OpenStack
+      class Real
+        def set_node_power_state(node_id, power_state)
```
TerryHowe: I didn't think these needed _node in them, the maintenance ones don't.

lib/fog/openstack/baremetal.rb
tzumainn: Thanks for the comments!  I'm happy to do the renaming, but I just wanted to check - it's not preferred to have the noun in the request name?  I had assumed that it was desired so that it was clear what the operation did; otherwise, we'll end up with requests/baremetal/set_power_state.rb, and it's not immediately clear what kind of object that operates on.

TerryHowe: Yeh, maybe they should have the noun in the request name and https://github.com/fog/fog/pull/3568 should change to have 'node_'

tzumainn:
@TerryHowe Let me know if you think the request name should be changed after all!  In the meantime I've updated the commit to remove the TODOs.  Thanks for reviewing!

TerryHowe:
I think the names with 'node' in them are better.  I like your approach better.

tzumainn:
@TerryHowe Great!  Would this pull request be mergeable then, or does it require further review?

TerryHowe:
I don't have merge privs, I just try to review OpenStack stuff.

plribeiro3000:
Bring it in! Thanks!

tzumainn:
Great - thanks!

geemus:
Thanks!

PR #3574, opened 2015-05-28
Fog::OpenStack::Compute: remove assumption about endpoint URL format

majewsky:
The reference implementation's endpoint URL ends with the tenant ID,
but that's not a requirement for the service to work. I work with
software that has differently-formatted endpoint URLs that don't
contain tenant IDs at all.

Review is much appreciated.

### DIFF
--- a/lib/fog/openstack/compute.rb
+++ b/lib/fog/openstack/compute.rb
@@ -368,7 +368,7 @@ module Fog
                 'Accept' => 'application/json',
                 'X-Auth-Token' => @auth_token
               }.merge!(params[:headers] || {}),
-              :path     => "#{@path}/#{@tenant_id}/#{params[:path]}",
+              :path     => "#{@path}/#{params[:path]}",
               :query    => params[:query]
             }))
           rescue Excon::Errors::Unauthorized => error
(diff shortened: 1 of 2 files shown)


mdarby:
Won't this conflict with existing implementations? i.e., those that do rely on a tenant id?

majewsky:
I thought so, too, at first, but all relevant operations get the tenant ID from elsewhere as far as I can see (sometimes implicitly through the token scope, sometimes as an explicit argument). I did not find a place where the `@tenant_id` variable was really used, besides that one mock implementation.

I can only say that our OpenStack implementation does not have the tenant ID in the endpoint URL, and it seems to work just fine for everything which we have implemented at the moment. (The OpenStack API is positively huge, but if the tenant ID in the endpoint URL were that relevant, we would have hit some cases already where our model would fail.)

TerryHowe:
I'd like to wait for more comments, but I too was thinking this would be okay for existing implementations.

majewsky:
Any update here? It would be really great to have this in the 1.31 release if there are no issues.

plribeiro3000:
@TerryHowe @mdarby Do you guys think this is ok to be merged?

Should we ask for tests?

TerryHowe:
However @mdarby feels on this is fine with me.

mdarby:
If Travis passes I'm sure we'd be fine. It seems a change like this (if it were to break the build) it would make it really :boom:.

geemus:
Alright, bringing it in, hopefully it will be smooth on release, but if not we'll sort it out.

geemus:
Thanks!

PR #3575, opened 2015-05-29
Upgraded fog-core to version 1.30

zephyr-dev:
We are using Fog in our application and are trying to eliminate the verbose warnings that older versions of Excon output. Since the master branch of Fog has fog-core locked to 1.27, the locked version of Excon is too low and the bug is present. We have upgraded the fog-core dependency to 1.30 in order to upgrade Excon.


geemus:
Thanks!

PR #3577, opened 2015-06-01
Fix OpenStack compute docs

TerryHowe:
The OpenStack compute docs had params for the request layer
in a model example.  I switched it to all model.


plribeiro3000:
Thanks!

geemus:
Thanks!

PR #3582, opened 2015-06-05
[openstack] Various fixes for Identity V3 API & related tests

dhague:


dhague:
Not sure why Travis complains - it was fine with the same commit in my repo: https://travis-ci.org/dhague/fog/builds/65906424. I'm guessing test instability rather than a true failure.

dhague:
Travis complains with:
Fog::Compute::when ecloud is passed with no available credentials#test_0001_returns ArgumentError [/home/travis/build/fog/fog/spec/fog/compute_spec.rb:13]:
ArgumentError expected but nothing was raised.

The thing is, even if I "git reset --hard" back to the 1.31.0 release I get the same error. I am guessing that the build is failing because of a change in some other gem.

plribeiro3000:
@dhague Can you force the version of `fog-ecloud` to `0.1.1`?

dhague:
@plribeiro3000 on it :)

dhague:
Build #5158.2 fails in tests/openstack/models/compute/security_group_tests.rb - exactly the same failure as #5144.7 in the 1.31.0 master release: https://travis-ci.org/fog/fog/jobs/65499301. In other words, just one of those from-time-to-time random failures...

plribeiro3000:
Ok, so we got an issue in the latest version of `fog-ecloud`.

I will see about that.

dhague:
Build is now happy :-)

TerryHowe:
There is a lot here, but in general, lgtm.

dhague:
Sorry about that, Terry - it was during the period where we were using & improving the code, but before it got pulled into Fog. We are now using a local clone of Fog instead, so any future changes will show up as individual commits which should be easier to review.
There is one significant change, which is that I removed the defaulting of the project domain that you added:
https://github.com/dhague/fog/commit/444340641d7ced972f711f7cdd723483a467808a#diff-3aa267766daa7f21447cd3c77b43bf39L353
This is because having such a default made it difficult to work with unscoped (i.e. non-project) tokens.

I am working on a better way of handling that at the moment - namely, only set the domain to "default" if a tenant is specified (i.e. we are effectively supporting a legacy ID v2 API case), but if project is specified then do not use a default domain. I will also address @clarsonneur's concerns about handling user domain & project domain in the cases where user/project are specified by name rather than by ID. Patch will appear here soon...

dhague:
<sigh> Ruby 1.8.7 again. On it...

geemus:
Thanks!

PR #3585, opened 2015-06-08
[openstack] create_volume: add metadata option

d063130:
analogous to [create_server](https://github.com/fog/fog/blob/master/lib/fog/openstack/requests/compute/create_server.rb#L14) the metadata option can be also used at create_volume (see [api-ref-v1](http://developer.openstack.org/api-ref-blockstorage-v1.html#createVolume) and [api-ref-v2](http://developer.openstack.org/api-ref-blockstorage-v2.html#createVolume))


mdarby:
lgtm

TerryHowe:
lgtm

geemus:
Thanks!

PR #3591, opened 2015-06-16
OpenStack port: add security groups to model and update request

d063130:
The security groups option is already there in the [create-request](https://github.com/fog/fog/blob/master/lib/fog/openstack/requests/network/create_port.rb#L13), it can also be used in the update_request (see [api-ref](http://developer.openstack.org/api-ref-networking-v2.html#updatePort)) and be shown in the related model

### DIFF
--- a/lib/fog/openstack/models/network/port.rb
+++ b/lib/fog/openstack/models/network/port.rb
@@ -15,6 +15,7 @@ module Fog
         attribute :device_owner
         attribute :device_id
         attribute :tenant_id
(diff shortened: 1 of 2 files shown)


TerryHowe:
Looks good, but Travis disagrees.

d063130:
sorry, I tested only the Real Class and not the mocked one

TerryHowe:
lgtm!

geemus:
Thanks!

PR #3593, opened 2015-06-17
[openstack] Add missing reference to openstack_project_name & simplify test code

dhague:
Critical bugfix - without this, authentication is domain-scoped even if project_name is specified.

### DIFF
--- a/spec/fog/openstack/identity_v3_spec.rb
+++ b/spec/fog/openstack/identity_v3_spec.rb
@@ -6,7 +6,7 @@ require 'vcr'
 
 RSpec.describe Fog::Identity::OpenStack::V3 do
 
-  before :each do |example|
+  before :all do |example|
     @os_auth_url = ENV['OS_AUTH_URL']
 
     # if OS_AUTH_URL is set but FOG_MOCK is not, don't record anything and just pass through the requests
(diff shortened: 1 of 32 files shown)


spec/fog/openstack/identity_v3_spec.rb
```diff
@@ -40,20 +40,18 @@
...
-            :openstack_auth_url => "#{@os_auth_url}/auth/tokens") unless @id_v3
-      end
+    VCR.use_cassette('idv3') do
+      @id_v3 = Fog::Identity::OpenStack::V3.new(
+          :openstack_project_name => ENV['OS_PROJECT_NAME'] || 'admin',
+          :openstack_domain_name => ENV['OS_USER_DOMAIN_NAME'] || 'Default',
+          :openstack_api_key => ENV['OS_PASSWORD'] || 'password',
+          :openstack_username => ENV['OS_USERNAME'] || 'admin',
+          :openstack_region => ENV['OS_REGION_NAME'] || 'RegionOne',
+          :openstack_auth_url => "#{@os_auth_url}/auth/tokens") unless @id_v3
```
TerryHowe: Unrelated to your change, I often think /auth/tokens should automatically be appended to the auth_url so Fog would be more compatible with standard OpenStack env configs.

geemus:
Thanks!

PR #3594, opened 2015-06-18
OpenStack Compute list methods unified interface

Ladas:

etc.

### DIFF
--- a/lib/fog/openstack/models/compute/addresses.rb
+++ b/lib/fog/openstack/models/compute/addresses.rb
@@ -7,9 +7,11 @@ module Fog
       class Addresses < Fog::Collection
         model Fog::Compute::OpenStack::Address
 
-        def all
-          load(service.list_all_addresses.body['floating_ips'])
+        def all(options = {})
+          load(service.list_all_addresses(options).body['floating_ips'])
         end
+
+        alias_method :summary, :all
 
         def get(address_id)
           if address = service.get_address(address_id).body['floating_ip']
--- a/lib/fog/openstack/compute.rb
+++ b/lib/fog/openstack/compute.rb
@@ -178,6 +178,7 @@ module Fog
 
       # Volume
       request :list_volumes
+      request :list_volumes_detail
       request :create_volume
       request :get_volume_details
       request :delete_volume
@@ -185,8 +186,10 @@ module Fog
       request :detach_volume
       request :get_server_volumes
 
+      # Snapshot
       request :create_volume_snapshot
       request :list_snapshots
+      request :list_snapshots_detail
       request :get_snapshot_details
       request :delete_snapshot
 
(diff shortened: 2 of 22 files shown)


Ladas:
@geemus hello, I think this cleanup was long needed, I am seeing like 4-5 different interfaces. It causes problems when we want to solve pagination and filtering in a generic way. I will send PR for each OpenStack service, so it's smaller chunks for review and we can revert just one service, if something goes wrong.

Let me know if it all makes sense.

TerryHowe:
lgtm.  Hope those deprecation messages aren't there for long.

lib/fog/openstack/models/compute/addresses.rb
```diff
@@ -7,10 +7,12 @@ class OpenStack
       class Addresses < Fog::Collection
         model Fog::Compute::OpenStack::Address
 
-        def all
-          load(service.list_all_addresses.body['floating_ips'])
+        def all(options = {})
+          load(service.list_all_addresses(options).body['floating_ips'])
         end
 
+        alias_method :details, :all
```
geemus: Could you give some context around the need for the new alias here?

geemus:
I think I see where this is headed. Alternatively, would it be less disruptive to leave all more like it is and add a `summary` method or something for the less detailed version? It seems at a glance at least that it would be less of a behavior change this way around. What do you think?

Ladas:
@geemus I picked details method, since it was already used in many places, so this way we will need more delegations on details method, but less deprecation warnings for existing details method.

I think as next step the delegation could go to super class, so we would define abstract interface of collection and :details method would delegate to :all in the super class. What do you think?

Not sure where the abstract interface should go, Fog::Collection is probably too generic, though would be nice to have same the interface in whole Fog for collections. So starting e.g. Fog::Collection::Cloud or Fog::Collection::Openstack ? Should it go to Fog::Core?

Same would go to Fog::Model::Openstack later, as I see it has the same problem with consistency.

Ladas:
@geemus by aggregating all the methods up there, I think the basic set of methods should be

all(options = {}), details(options = {}), get(uuid), destroy(uuid)

maybe methods find_by_id and find_by_name, or theoretically generic method_missing finder

Ladas:
@geemus about the summary vs details method. I see about 5 occurrences of all(detailed=true) and about 7 of details method. So I guess it will mean about the same amount of deprecation warnings, whatever way we pick. :-)

geemus:
Hmm. I think in other providers all is usually a detailed view, so maybe we should stick to that if it is a comparable amount of work/deprecation?

Ladas:
@geemus yeah, I guess it's less work for me :-)

geemus:
Sounds good, :+1:

geemus:
I think we reached some conclusion on this and are moving in the right direction. Looks like this is no longer a clean merge though, could you take a look and figure out what needs done here? Thanks!

Ladas:
@geemus yes, there was a conflict in empty line :-)

should be ready now

geemus:
At least it was an easy conflict resolution, thanks!

geemus:
Thanks!

PR #3595, opened 2015-06-18
OpenStack Baremetal list methods unified interface

Ladas:

etc.

### DIFF
--- a/lib/fog/openstack/requests/baremetal/list_chassis.rb
+++ b/lib/fog/openstack/requests/baremetal/list_chassis.rb
@@ -2,18 +2,12 @@ module Fog
   module Baremetal
     class OpenStack
       class Real
-        def list_chassis(parameters=nil)
-          if parameters
-            query = parameters.each { |k, v| parameters[k] = URI::encode(v) }
-          else
-            query = {}
-          end
-
+        def list_chassis(options = {})
           request(
             :expects => [200, 204],
             :method  => 'GET',
             :path    => 'chassis',
-            :query   => query
+            :query   => options
           )
         end
       end # class Real
--- a/lib/fog/openstack/requests/baremetal/list_drivers.rb
+++ b/lib/fog/openstack/requests/baremetal/list_drivers.rb
@@ -2,17 +2,18 @@ module Fog
   module Baremetal
     class OpenStack
       class Real
-        def list_drivers
+        def list_drivers(options = {})
           request(
             :expects => [200, 204],
             :method  => 'GET',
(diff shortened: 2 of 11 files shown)


TerryHowe:
I'm not sure what standard we are going for across OpenStack here, but looks fine to me.

lib/fog/openstack/requests/baremetal/list_chassis.rb
```diff
@@ -2,18 +2,14 @@ module Fog
...
     class OpenStack
       class Real
-        def list_chassis(parameters=nil)
-          if parameters
-            query = parameters.each { |k, v| parameters[k] = URI::encode(v) }
-          else
-            query = {}
-          end
+        def list_chassis(options = {})
+          options = {} unless options
```
geemus: I don't think this line should be needed, given the default argument?

lib/fog/openstack/requests/baremetal/list_drivers.rb
```diff
@@ -2,17 +2,18 @@ module Fog
   module Baremetal
     class OpenStack
       class Real
-        def list_drivers
+        def list_drivers(options = {})
           request(
```
geemus: If, on the other hand, the override is needed, this is missing it.

geemus:
Couple small comments around defaulting/values. Otherwise looks good (I also didn't see any deprecation warnings, but I don't think it is needed in this case if I read things correctly).

lib/fog/openstack/requests/baremetal/list_chassis.rb
Ladas: ok, will remove it

lib/fog/openstack/requests/baremetal/list_drivers.rb
Ladas: not sure I understand what is missing?

lib/fog/openstack/requests/baremetal/list_chassis.rb
Ladas: done, also on other places

lib/fog/openstack/requests/baremetal/list_drivers.rb
Ladas: ah you are talking about options = {} unless options, I removed it from all places. It shouldn't be needed

geemus: Yep, thanks!

geemus:
Thanks!

PR #3596, opened 2015-06-19
OpenStack Keystone list methods unified interface

Ladas:

Also filtering of options has been deleted, so Fog is now
passing everything, this way we don't have to support changes
in API. Also default values page and per_page has been deleted
because these default values are already set on keystone side.
Default values has been deleted also from recorded VCRs.

etc.

### DIFF
--- a/lib/fog/openstack/requests/identity_v3/list_endpoints.rb
+++ b/lib/fog/openstack/requests/identity_v3/list_endpoints.rb
@@ -3,25 +3,18 @@ module Fog
     class OpenStack
       class V3
         class Real
-          def list_endpoints(options={})
-            params = Hash.new
-            params['service_id'] = options[:service_id] if options[:service_id]
-            params['interface'] = options[:interface] if options[:interface]
-
-            params['page'] = options.fetch(:page, 1)
-            params['per_page'] = options.fetch(:per_page, 30)
-
+          def list_endpoints(options = {})
             request(
-                :expects => [200],
-                :method => 'GET',
-                :path => "endpoints",
-                :query => params
+              :expects => [200],
+              :method  => 'GET',
+              :path    => "endpoints",
+              :query   => options
             )
           end
         end
 
         class Mock
-          def list_endpoints
+          def list_endpoints(options={})
 
           end
         end
--- a/lib/fog/openstack/requests/identity_v3/list_role_assignments.rb
+++ b/lib/fog/openstack/requests/identity_v3/list_role_assignments.rb
@@ -3,29 +3,35 @@ module Fog
     class OpenStack
       class V3
         class Real
-          def list_role_assignments(options={})
-            params = Hash.new
-            params['group.id'] = options[:group_id] if options[:group_id]
-            params['role.id'] = options[:role_id] if options[:role_id]
-            params['scope.domain.id'] = options[:domain_id] if options[:domain_id]
-            params['scope.project.id'] = options[:project_id] if options[:project_id]
-            params['user.id'] = options[:user_id] if options[:user_id]
-            params['effective'] = options[:effective] if options[:effective]
-
-            params['page'] = options.fetch(:page, 1)
-            params['per_page'] = options.fetch(:per_page, 30)
+          def list_role_assignments(options = {})
+            # Backwards compatibility name mapping
+            name_mapping = {
+              :group_id   => 'group.id',
+              :role_id    => 'role.id',
+              :domain_id  => 'scope.domain.id',
+              :project_id => 'scope.project.id',
+              :user_id    => 'user.id',
+            }
+            name_mapping.keys.each do |key|
+              if (opt = options.delete(key))
+                Fog::Logger.deprecation("Calling OpenStack[:keystone].list_role_assignments(options) with deprecated"\
+                                        " option '#{key}'. Use option '#{name_mapping[key]}' which is defined in"\
+                                        " keystone documentation.")
+                options[name_mapping[key]] = opt
+              end
+            end
 
             request(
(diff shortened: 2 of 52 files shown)


Ladas:
@geemus this one got bigger, I am getting rid of filtering of options we send and default values. This was not done in other services much. At least this way, we don't have to support all changes in APIs

I deleted the default values from VCRs. I kinda like the VCRs, would be nice to use them for whole openstack specs.

What do you think?

lib/fog/openstack/requests/identity_v3/list_endpoints.rb
```diff
@@ -3,29 +3,22 @@ module Identity
     class OpenStack
       class V3
         class Real
-          def list_endpoints(options={})
-            params = Hash.new
-            params['service_id'] = options[:service_id] if options[:service_id]
-            params['interface'] = options[:interface] if options[:interface]
-
-            params['page'] = options.fetch(:page, 1)
-            params['per_page'] = options.fetch(:per_page, 30)
```
Ladas: @geemus we should not default values here, when they are already defaulted on keystone side, right? One source of defaults should be enough

lib/fog/openstack/requests/identity_v3/list_endpoints.rb
```diff
@@ -3,29 +3,22 @@ module Identity
     class OpenStack
       class V3
         class Real
-          def list_endpoints(options={})
-            params = Hash.new
-            params['service_id'] = options[:service_id] if options[:service_id]
```
Ladas: I wonder how to approach this in general. For some reason I would rather have if it passes all I send it, and i will have just exception back, when bad attr will be send. This way you just have to look in Openstack API documentation

If we decide to filter it, it can be done more effectively (line on one line), similar too what we do on create/update with vanilla options. Nevertheless all filtering should at least log warning with filtered out options, right? Otherwise it's hard to debug if it just becomes lost.

lib/fog/openstack/requests/identity_v3/list_role_assignments.rb
```diff
@@ -3,33 +3,39 @@ module Identity
...
-            params['scope.domain.id'] = options[:domain_id] if options[:domain_id]
-            params['scope.project.id'] = options[:project_id] if options[:project_id]
-            params['user.id'] = options[:user_id] if options[:user_id]
-            params['effective'] = options[:effective] if options[:effective]
-
-            params['page'] = options.fetch(:page, 1)
-            params['per_page'] = options.fetch(:per_page, 30)
+          def list_role_assignments(options = {})
+            # Backwards compatibility name mapping
+            name_mapping = {
```
Ladas: For some reason, this one method contains totally different attributes, than rest of keystone methods. 

I wonder if Fog should just offer the single pane of glass here. I could just remove the deprecation and support both standard option names and those that are in API doc. I guess we should be solving such inconsistencies in Fog, or not? It definitely puts more weight on Fog's shoulders. :-)

Ladas:
hm weird jruby 1.9 failure 

  Fog::Compute[:openstack] | server requests (openstack) +++++      
      tests/openstack/requests/compute/server_tests.rb
        success
        Fog::Compute[:openstack] | server requests (openstack)
      - succeeds
        expected => true
        returned => false

Ladas:
Now the same bug, bug in ruby 2.1.1

  Fog::Compute[:openstack] | server requests (openstack) +++++      
      tests/openstack/requests/compute/server_tests.rb
        success
        Fog::Compute[:openstack] | server requests (openstack)
      - succeeds
        expected => true
        returned => false

Ladas:
the random error in CI appears here, so probably a sorting issue?

tests("#block_device_mapping_v2").succeeds do
  compute.servers.get(@server_id).volumes.collect(&:id).sort == [@volume1_id, @volume2_id].sort
end

lib/fog/openstack/requests/identity_v3/list_role_assignments.rb
dhague: The reason for this is that in this API call (according to http://developer.openstack.org/api-ref-identity-v3.html#listRoleAssignments) the parameter names have dots in them, which is not particularly Ruby-friendly. In the rest of the Identity V3 API the parameter names usually have underscores instead of dots.

dhague:
Looks good to me (for the Identity v3 stuff, at least) - code is simpler, and the tests pass without modification.

lib/fog/openstack/requests/identity_v3/list_endpoints.rb
geemus: Probably true, yeah.

lib/fog/openstack/requests/identity_v3/list_endpoints.rb
geemus: True, maybe just sending everything through would be cleaner/simpler.

lib/fog/openstack/requests/identity_v3/list_role_assignments.rb
geemus: Gross. It seems like something we should handle if it needs handling (requiring end users to figure it out seems bad.

geemus:
Thanks!

lib/fog/openstack/requests/identity_v3/list_role_assignments.rb
Ladas: @geemus ok, I will just remove the deprecation and we will support both formats.

@dhague yeah I saw that and I was like :-O, devs had to think it will be fun to change consistency just for this one command :-)

PR #3600, opened 2015-06-22
OpenStack Glance and Heat list methods unified interface

Ladas:
of the Hash goes directly to request :query. ArgumentError
is raised when mandatory param is missing.

etc.

### DIFF
--- a/lib/fog/openstack/models/image/images.rb
+++ b/lib/fog/openstack/models/image/images.rb
@@ -7,12 +7,18 @@ module Fog
       class Images < Fog::Collection
         model Fog::Image::OpenStack::Image
 
-        def all
-          load(service.list_public_images_detailed.body['images'])
+        def all(options = {})
+          load(service.list_public_images_detailed(options).body['images'])
         end
 
-        def details(attribute=nil, query=nil)
-          load(service.list_public_images_detailed(attribute, query).body['images'])
+        def summary(options = {})
+          load(service.list_public_images(options).body['images'])
+        end
+
+        def details(options = {}, deprecated_query = nil)
+          Fog::Logger.deprecation("Calling OpenStack[:glance].images.details will be removed, "\
+                                  " call .images.all for detailed list.")
+          load(service.list_public_images_detailed(options, deprecated_query).body['images'])
         end
 
         def find_by_id(id)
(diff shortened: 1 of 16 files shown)


geemus:
I thought we were leaning the other direction now, ie having all be detailed and #summary (or similar) be non-detailed? Seems like this is in the opposite direction, so maybe I misunderstood.

Ladas:
@geemus ah, I thought we agreed on opposite, at least all the other patches has been merged with 'details' methods and all as a summary

Ladas:
@geemus ah, yeah you wrote it the other way in #3594. Though I wrote it's less work when I will not change it from having methods 'all' and 'details', I thought you want it like this since the 2 PRs were merged. :-) Ok, so should I change it all to having 'all' and 'summary'? :-)

I see that other providers doesn't have details nor summary method, I wonder if we even need the summary. I guess it make sense then to change all to be detailed, so we hold the cross provider interface

geemus:
Sorry for the confusion, I guess I overlooked some of the specifics in trying to get through the review queue. Yeah, although within openstack it may have been as you describe, in the broader fog #all tends to be detailed. We could probably skip the summary versions altogether, at least for the time being. If you have details you usually don't need summary (outside of performance/optimization perhaps).

Ladas:
@geemus no problem
let's just hold this one before merging please. I brought here heat detailed list, that is actually not even documented, so I need to test it against real env.

I'll let you know when this passes

geemus:
Sounds good, thanks!

Ladas:
@geemus I guess we can make summary as part of standard interface. Though I haven't seen the need in the other providers. In OpenStack, maybe like 30% of the API calls supports it. Yeah, for example for Heat, it can make difference for the performance, since each record can have large data in the detailed list

geemus:
Yeah. Perhaps we can just add it as needed (and where available). Otherwise we may be able to just alias to all/details (as I think you've already started to do) so that it can still work even when summary isn't available (or isn't yet available).

Ladas:
@geemus yeah, I've added the aliases to all places. I will send abstract interface class to fog-core code. Then I guess the alias could be there as default (since the abstract interface class doesn't have to be that strict :-)) and I could remove it from all the classes and just keep it in the superclass.

geemus:
True, sounds like a plan.

Ladas:
@geemus it should be ready now

the detailed list is not fully working, so I commented it with todo, pointing to upstream bug. We can switch to that once it's fully functional.

Also the existing test data doesn't seem to be reflecting reality much. I might revisit that in the future, or I will just replace everything with VCRs. Having VCRs will show, that the tests ran against real system at least once.

geemus:
Sounds good, thanks!

PR #3602, opened 2015-06-23
OpenStack Keystone and Ironic change all to detailed list

Ladas:
Changing 'all' to 'detailed' list and adding summary method for quick
non detailed lists.

This change is backwards compatible since all method will return
more data. And the alias details methods has not been released yet.


Ladas:
@geemus ok so changing to 'all' and 'summary' for the merged PRs. The before released details methods are just deprecated. And the aliases to details were not released yet, so no need to make them backwards compatible right?

geemus:
Yep, sounds good. Thanks for working through this with me.

geemus:
Thanks!

PR #3604, opened 2015-06-23
OpenStack Cinder, Swift and Tuskar list methods unified interface

Ladas:

etc.


geemus:
Thanks!

PR #3607, opened 2015-06-24
add extend_volume request to Fog::Volume::OpenStack

majewsky:
I would've liked to include tests, but the mock service is not up to it.
I cannot even create a volume and retrieve it again using the same ID.

### DIFF
--- a/lib/fog/openstack/requests/volume/extend_volume.rb
+++ b/lib/fog/openstack/requests/volume/extend_volume.rb
@@ -8,7 +8,7 @@ module Fog
             :expects  => 202,
             :method   => 'POST',
             :path     => "volumes/#{volume_id}/action",
-            :body     => Fog::JSON.encode(body),
+            :body     => Fog::JSON.encode(body)
           )
         end
       end
--- a/lib/fog/openstack/requests/volume/extend_volume.rb
+++ b/lib/fog/openstack/requests/volume/extend_volume.rb
@@ -1,0 +1,25 @@
+module Fog
+  module Volume
+    class OpenStack
+      class Real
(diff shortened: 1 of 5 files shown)


geemus:
@Ladas @mdarby could you guys take a peak? Thanks!

dhague:
I'll have a look with @majewsky tomorrow to see about using VCR-based tests for this (we are both at SAP, albeit in different countries)

geemus:
@dhague sounds great, just let me know if you need anything. Thanks!

TerryHowe:
looks good, but more tests would always be nice.

Ladas:
looks good, the VCR tests would be great, at least some form of tests should be present

geemus:
Thanks for taking looks, just let me know once another pass has been taken and I'll review anew. Thanks!

majewsky:
I rebased on master and added tests.

lib/fog/openstack/requests/volume/extend_volume.rb
```diff
@@ -0,0 +1,25 @@
...
+  module Volume
+    class OpenStack
+      class Real
+        def extend_volume(volume_id, size)
+          body = { 'os-extend' => { 'new_size' => size } }
+          request(
+            :expects  => 202,
+            :method   => 'POST',
+            :path     => "volumes/#{volume_id}/action",
+            :body     => Fog::JSON.encode(body),
```
Ladas: ruby 1.8.7 has problems with trailing commas

majewsky:
I really understand now why you would want to drop 1.8.7 support. :)

geemus:
Yeah, it can be a real pain. Thanks!

PR #3608, opened 2015-06-24
support neutron networks labled as public

hh:
Some folks use directly connected neutron networks that are actually public networks.
This patch supports using both floating_ip_addresses AND neutron networks with 'public' in the name.

### DIFF
--- a/lib/fog/openstack/models/compute/server.rb
+++ b/lib/fog/openstack/models/compute/server.rb
@@ -135,16 +135,24 @@ module Fog
           all_floating.empty? ? manual : all_floating
         end
 
-        alias_method :public_ip_addresses, :floating_ip_addresses
+        def public_ip_addresses
+          if floating_ip_addresses.empty?
+            addresses.find_all{|s| s[0] =~ /public/i}.map{|a| a[1][0]['addr']}
+          else
+            floating_ip_addresses
+          end
+        end
 
         def floating_ip_address
           floating_ip_addresses.first
         end
 
-        alias_method :public_ip_address, :floating_ip_address
+        def public_ip_address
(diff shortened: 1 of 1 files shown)


lib/fog/openstack/models/compute/server.rb
```diff
@@ -135,16 +135,24 @@ def floating_ip_addresses
           all_floating.empty? ? manual : all_floating
         end
 
-        alias_method :public_ip_addresses, :floating_ip_addresses
+        def public_ip_addresses
+          if floating_ip_addresses.empty?
+            addresses.find_all{|s| s[0] =~ /public/i}.map{|a| a[1][0]['addr']}
```
TerryHowe: Some tests here would sure help me understand the various use cases.

hh: ```public_ip_addresses``` are not the same as ```floating_ip_addresses```

If you have attached floating ip addresses to the vm, you can assume that those are public.

If you don't use floating_ip_addresses and instead have networks with /public/i in the name, use those instead.

Otherwise the logic is the same.

I'm not yet familiar with the testing framework, but with some quick direction / pointers I could try to code something up.

geemus:
Should be some examples here: https://github.com/fog/fog/blob/b2c6e0df30eae7fb9154ba6bd340f27c0c158855/tests/openstack/models/compute/server_tests.rb (though tests around requests tend to be better than around models, at least in part because they mostly expect to call through to requests anyway). Hope that helps you get closer, but certainly happy to discuss/help where I can.

plribeiro3000:
@hh Could you open this against `fog-openstack` isntead?

Thanks!

PR #3609, opened 2015-06-26
first bits of a spec for Fog::OpenStack::Volume

majewsky:
This has some code duplication with the Fog::Identity::OpenStack::V3 spec right now. I'm going to discuss this with @dhague next week and consolidate this.

Nonetheless, I'd like to have this merged to be able to move forward with adding test coverage for #3607.


geemus:
Thanks!

PR #3611, opened 2015-06-29
[cloudstack] Use `isdefault` in `listNetworks` response

yyuu:
Add support for `isdefault` field in the response from CloudStacks `listNetworks` request.

https://cloudstack.apache.org/api/apidocs-4.4/user/listNetworks.html


geemus:
Thanks!

PR #3612, opened 2015-06-29
Missing network init uri

Ladas:
Missing network init uri, got lost probably in cleanup in
https://github.com/fog/fog/commit/83b61986bd2787dd22dc8967f28d0c78d8752aff

### DIFF
--- a/lib/fog/openstack/core.rb
+++ b/lib/fog/openstack/core.rb
@@ -161,16 +161,15 @@ module Fog
           @openstack_must_reauthenticate = false
           @auth_token = credentials[:token]
           @openstack_management_url = credentials[:server_management_url]
-          uri = URI.parse(@openstack_management_url)
         else
           @auth_token = @openstack_auth_token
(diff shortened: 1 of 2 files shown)


Ladas:
@dhague you probably forgot it while doing the cleanup https://github.com/fog/fog/commit/83b61986bd2787dd22dc8967f28d0c78d8752aff

I am getting exception with missing uri

dhague:
Good catch, thanks - might it be better to put that line into the initialize_identity method so that it's picked up for other openstack services?

Ladas:
@dhague yeah I guess we could set it in authenticate method, where the code is now

so e.g. @openstack_management_uri = URI.parse(@openstack_management_url) ?

Ladas:
@dhague changed, something like this was on your mind?

dhague:
That will do the job nicely :)

geemus:
Thanks!

PR #3613, opened 2015-06-29
Don't remove nil keys in OpenStack#create_subnet

mdarby:
The [docs](http://developer.openstack.org/api-ref-networking-v2.html) state that a `gateway_ip` key may be passed in with nil to represent "no gateway_ip" so that a gateway IP address isn't automatically assigned.

Discovered via: http://stackoverflow.com/questions/30930343/nil-not-turning-into-null-in-request-on-fog

### DIFF
--- a/lib/fog/openstack/requests/network/create_subnet.rb
+++ b/lib/fog/openstack/requests/network/create_subnet.rb
@@ -7,22 +7,22 @@ module Fog
             'subnet' => {
               'network_id' => network_id,
               'cidr'       => cidr,
-              'ip_version' => ip_version,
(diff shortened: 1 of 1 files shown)


mdarby:
Can we get this merged @geemus? :)

TerryHowe:
lgtm!

geemus:
Certainly, thanks!

abattye:
There seem to be a few issues with this running against the Kilo neutron API. In our tests The API rejects nil/None values for enable_dhcp, allocation_pools and tenant_id.  We'll do some further testing and try and pull a update together to try and address this.

PR #3614, opened 2015-06-30
OpenStack missing list unification of Network and Volume

Ladas:
Missing summary methods under network. Missing common interface
in list_snapshots and detailed method under volume.


TerryHowe:
lgtm

geemus:
Thanks@

Issue #3541, opened 2015-04-23
Rackspace snapshots - Force = true ignored


yannh:
So this call:

snapshot = volume.create_snapshot :display_name => snapshot_name, :force => true, :volume_id => volume.id

Fails with the following stack:

        from /var/lib/gems/1.9.1/gems/excon-0.45.3/lib/excon/middlewares/response_parser.rb:8:in `response_call'
        from /var/lib/gems/1.9.1/gems/excon-0.45.3/lib/excon/connection.rb:373:in `response'
        from /var/lib/gems/1.9.1/gems/excon-0.45.3/lib/excon/connection.rb:237:in `request'
        from /var/lib/gems/1.9.1/gems/fog-core-1.30.0/lib/fog/core/connection.rb:81:in `request'
        from /var/lib/gems/1.9.1/gems/fog-1.29.0/lib/fog/rackspace/service.rb:42:in `request'
        from /var/lib/gems/1.9.1/gems/fog-1.29.0/lib/fog/rackspace/block_storage.rb:89:in `request'
        from /var/lib/gems/1.9.1/gems/fog-1.29.0/lib/fog/rackspace/requests/block_storage/create_snapshot.rb:43:in `create_snapshot'
        from /var/lib/gems/1.9.1/gems/fog-1.29.0/lib/fog/rackspace/models/block_storage/snapshot.rb:65:in `save'
        from /var/lib/gems/1.9.1/gems/fog-core-1.30.0/lib/fog/core/collection.rb:51:in `create'
        from /var/lib/gems/1.9.1/gems/fog-1.29.0/lib/fog/rackspace/models/block_storage/volume.rb:91:in `create_snapshot'
        from ./snapshot.rb:65:in `<main>'

If we go up that trace:

volume.create_snapshot receives the options hash with :force set to true. It calls create in /lib/fog/core/collection.rb passing it the options - so in create, attributes[:force] is set to true.

Create then calls object.save in rackspace/models/block_storage/snapshot.rb without passing it any argument. Except that this save function, there is a parameter "force" that defaults to false. 

Long story short: it's not currently possible to take a snapshot of a disk that is connected to a VM using fog. 

I am trying to think of a fix but it won't be easy to fix without breaking backward compatibility...

yannh:
Considering the fact that core/collection.rb calls save without parameters, and touching core is probably not what we want to do, adding an attribute :force to the Snapshot class seems inevitable to me.

Adding a :force attribute to the snapshot class and 

[...]

PR #3615, opened 2015-06-30
Send :force flag in Snapshot payload

mdarby:
Now sending the `:force` argument as per the [docs](http://developer.openstack.org/api-ref-blockstorage-v1.html#createSnapshot)

(This references #3541)

``` ruby
bs = Fog::Rackspace::BlockStorage.new
v = bs.volumes.create({size: 75, display_name: "FOO"})
...
s = bs.snapshots.create({display_name: "BAR", volume_id: v.id, force: true})
```


geemus:
Thanks!

PR #3616, opened 2015-07-01
Openstack security groups list typo

Ladas:
Openstack security groups list typo with elseif


Ladas:
@geemus bad typo, do you think we can merge before release?

geemus:
Oof, yeah that would be problematic.

Ladas:
@geemus that was the last and our internal tests are passing \o/

Though it also doesn't cover everything. I will need to find time and do 100% requests and models tests coverage using VCR in Fog. That should catch all the stupid bugs. :-)

geemus:
Definitely, sounds good. I believe tomorrow should be release day also, fwiw.

PR #3619, opened 2015-07-06
[openstack] reduce code duplication between specs

majewsky:
This adds an `RSpec.shared_context` which contains the code that the `before :all` blocks of both specs need. The setup logic of both specs has been merged and simplified.

I also used the opportunity to document the workings of these specs at the start of `shared_context.rb`, to encourage the other collaborators to create their own specs for the various other OpenStack services.


majewsky:
@dhague You gave a quick thumbs up on the phone earlier today. Do you want to take a closer look before this is merged?

dhague:
On closer review, it still looks good to me. :+1:

geemus:
Thanks!

PR #3623, opened 2015-07-07
process runcmd when creating a customspec  from cloudinit in vsphere

karmab:
this allows input to be compliant with cloudinit syntax when passing first run commands 
so the following input could be used instead of very specific   
runcmd:
 - echo superbien > /tmp/prout
 - echo 10.32.244.60 karmasat61.xxx >> /etc/hosts

(of course, for some specific items, typically windows bits, the vsphere naming can still be used)

### DIFF
--- a/lib/fog/vsphere/requests/compute/cloudinit_to_customspec.rb
+++ b/lib/fog/vsphere/requests/compute/cloudinit_to_customspec.rb
@@ -16,6 +16,7 @@ module Fog
           custom_spec['encryptionKey']    = user_data['encryptionKey'] if user_data.key?('encryptionKey')
           custom_spec['globalIPSettings'] = user_data['globalIPSettings'] if user_data.key?('globalIPSettings')
           custom_spec['identity']         = user_data['identity'] if user_data.key?('identity')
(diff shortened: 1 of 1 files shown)


geemus:
@nirvdrum could you take a peak at this one also? Thanks!

xc8tlik:
+1 for getting this merged.

It will facilitate running commands against nodes built by Foreman with `user_data` templates (["image provisioning without SSH"](http://www.theforeman.org/manuals/1.11/index.html#5.2.9VMwareNotes)) as per these [forum](http://serverfault.com/questions/699669/foreman-vm-provisioning-without-ssh-and-dhcp-dns) [posts](https://groups.google.com/forum/#!topic/foreman-users/xJf9b2hXJL4).

geemus:
Unfortunately, it looks like this is no longer a clean merge. Could I get a rebase? Thanks!

karmab:
so this needs to be done at fog-vsphere repo instead now. will do

karmab:
opened brand new https://github.com/fog/fog-vsphere/pull/34

geemus:
Thanks!

PR #3624, opened 2015-07-08
[openstack] add volume transfer models/requests

majewsky:
Includes some reduction of code duplication in the volume spec through introduction of helper methods.


TerryHowe:
lgtm

geemus:
Thanks!

PR #3625, opened 2015-07-08
Adding Openstack Collection base class

Ladas:
Several parts of this patch

1. base class documents how the standard interface should look
like. This doesn't bring anything to the code, since abstract
classes are redundant in ruby, but it is nice to see what should
be the standard interface.

2. load_response method saves the whole response. Usually not
only list of records is sent. It can contain also metatada about
pagination, or other useful metadata.

### DIFF
--- a/lib/fog/openstack/models/collection.rb
+++ b/lib/fog/openstack/models/collection.rb
@@ -1,0 +1,45 @@
+require 'fog/core/collection'
+
+module Fog
+  module OpenStack
+    class Collection < Fog::Collection
+      # It's important to store the whole response, it contains e.g. important info about whether there is another
+      # page of data.
+      attr_accessor :response
+
+      def load_response(response, index = nil)
+        # Delete it index if it's there, so we don't store response with data twice, but we store only metadata
+        objects = index ? response.body.delete(index) : response.body
+
+        clear && objects.each { |object| self << new(object) }
+        self.response = response
+        self
+      end
+
+      ##################################################################################################################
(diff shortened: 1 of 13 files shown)


Ladas:
@geemus hello, I joined several parts into this patch, I might divide it into more if needed. Not sure if this should be more general behavior, meaning it should go to fog-core. I guess as OpenStack specific it should stay here. What do you think?

let me know if it all makes sense.

Ladas:
btw. I will need to switch base class in all collection models, I can do that as next PR, just testing neutron

lib/fog/openstack/models/collection.rb
```diff
@@ -0,0 +1,47 @@
...
+
+module Fog
+  module OpenStack
+    class Collection < Fog::Collection
+      # It's important to store the whole response, it contains e.g. important info about whether there is another
+      # page of data.
+      attr_accessor :response
+      # In some cases it's handy to be able to store the tenant for the record, e.g. swift doesn't contain tenant info
+      # in the result, so we can track it in this attribute based on what tenant was used in the request
+      attr_accessor :tenant
```
TerryHowe: Can we call this project?  tenant is deprecated.

Ladas: ah, right, keystone v3 is using project. Will do, though I placed this badly, I need it in Fog::Model actually, to track project per record

TerryHowe:
lgtm

geemus:
Looks good, thanks!

PR #3627, opened 2015-07-09
Openstack Model base class

Ladas:
base class documents how the standard interface should look
like. This doesn't bring anything to the code, since abstract
classes are redundant in ruby, but it is nice to see what should
be the standard interface.

Base class also define the repeated methods initialize and save,
so we don't need to define them in subclasess.

project is storable to attribute, since not all results contain
project associated, it's good to to be able to store the project
used when processing the collection.

### DIFF
--- a/lib/fog/openstack/models/model.rb
+++ b/lib/fog/openstack/models/model.rb
@@ -1,0 +1,44 @@
+require 'fog/core/model'
+
+module Fog
+  module OpenStack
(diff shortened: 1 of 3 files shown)


Ladas:
@geemus and the same as for collection should be for model

@TerryHowe the project attr belongs here

TerryHowe:
lgtm

geemus:
Thanks!

PR #3629, opened 2015-07-15
Fog::OpenStack::Model base class for all openstack models

Ladas:
Fog::OpenStack::Model base class for all openstack models,
also deleting repeated initialize and save methods, since
they are defined in base class now.


TerryHowe:
lgtm

geemus:
Thanks!

PR #3630, opened 2015-07-15
Fog::OpenStack::Collection base class for all openstack collections

Ladas:
Fog::OpenStack::Collection base class for all openstack collections,
also replacing load method with load_response method, which stores
the whole response on Collection object, in the case that we need
other metadata e.g. for pagination.

Also removing :summary and :find_by_id aliases, since they are
defined in the base method.


TerryHowe:
lgtm

geemus:
Thanks!

PR #3633, opened 2015-07-16
[openstack] Add test for subnet creation

dhague:
Initially the test failed because too many params are set to nil instead of just gateway_ip (because of https://github.com/fog/fog/pull/3613) - in fact gateway_ip is the *only* parameter where a null value should be passed to OpenStack.


mdarby:
:+1:

geemus:
Thanks!

PR #3634, opened 2015-07-16
[openstack] Add tests for subtree_as_list and parents_as_list in hierarchical projects

dhague:
Also enhanced identity service to make the ID of the current user accessible as current_user_id.

Note that the returned lists are empty (even if *_as_ids calls return values) unless the calling user has a role in the projects - i.e. the only projects returned are those for which the user has a role.


TerryHowe:
+1

mdarby:
:+1:

geemus:
Thanks!

PR #3635, opened 2015-07-21
Allow OpenStack storage adapter to take :openstack_management_url option

tdg5:
It seems like the value used by `Fog::Storage::OpenStack::Real` for the `@openstack_management_url` is one of the few things that is not configurable on the adapter at initialization. This PR aims to make it configurable.

When paired with a valid `:openstack_auth_token` this should make it possible for the adapter to connect to Swift immediately without authenticating with Keystone.

This is useful in situations where many containers/tenants must be accessed by many processes, as it allows for caching auth information in a local file or shared memory or even something like redis. This is valuable in scenarios where keystone might not be able to handle the barrage of authentication requests for `m processes * n tenants`.

I'm open to subclassing `Fog::Storage::OpenStack::Real` if that's preferable to maintainers, but in my experimentation, the subclass I was able to create was missing some Fog magic like `adapter_instance.directories`.

One side note about this PR, presently, `:openstack_auth_token` issues a warning that it is unrecognized. I'm not sure if this is by intention or an accident, but to that end this PR makes `:openstack_auth_token` and `:openstack_management_url` recognized configuration options.

A final note, I looked around for tests that I could update, but didn't see anything that seemed relevant. Please point me to relevant tests if I have missed them!

Thanks!

### DIFF
--- a/lib/fog/openstack/storage.rb
+++ b/lib/fog/openstack/storage.rb
@@ -8,7 +8,8 @@ module Fog
       recognizes :persistent, :openstack_service_name,
                  :openstack_service_type, :openstack_tenant,
                  :openstack_region, :openstack_temp_url_key,
-                 :openstack_endpoint_type
+                 :openstack_endpoint_type, :openstack_auth_token,
+                 :openstack_management_url
 
       model_path 'fog/openstack/models/storage'
       model       :directory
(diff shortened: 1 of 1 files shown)


TerryHowe:
loodks good, just makes me wish auth was handled in one place and there was some mechanism to cache the service catalog.

geemus:
Thanks!

PR #3636, opened 2015-07-22
OpenStack compute add full support of availability zones

Ladas:
OpenStack compute add full support of availability zones


TerryHowe:
:+1:

geemus:
Thanks!

Issue #22, opened 2010-09-22
ec2 micro instances


matthewford:
Using the knife gem (chef) I get the following error when trying to launch a t1.micro instance:

```
knife ec2 server create "role[app]" -i ami-fae20893 -f t1.micro -x ubuntu -Z us-east-1a -k ~/.chef/clientserver.pem 
/home/matt/.rvm/gems/ruby-1.9.2-p0@rails3/gems/excon-0.2.2/lib/excon/connection.rb:69:in `request': UnsupportedOperation => AMIs with an instance-store root device are not supported for the instance type 't1.micro'. (Fog::AWS::EC2::Error)
    from /home/matt/.rvm/gems/ruby-1.9.2-p0@rails3/gems/fog-0.2.30/lib/fog/connection.rb:20:in `request'
    from /home/matt/.rvm/gems/ruby-1.9.2-p0@rails3/gems/fog-0.2.30/lib/fog/aws/ec2.rb:194:in `request'
    from /home/matt/.rvm/gems/ruby-1.9.2-p0@rails3/gems/fog-0.2.30/lib/fog/aws/requests/ec2/run_instances.rb:101:in `run_instances'
    from /home/matt/.rvm/gems/ruby-1.9.2-p0@rails3/gems/fog-0.2.30/lib/fog/aws/models/ec2/server.rb:141:in `save'
    from /home/matt/.rvm/gems/ruby-1.9.2-p0@rails3/gems/fog-0.2.30/lib/fog/collection.rb:47:in `create'
    from /home/matt/.rvm/gems/ruby-1.9.2-p0@rails3/gems/chef-0.9.8/lib/chef/knife/ec2_server_create.rb:124:in `run'
    from /home/matt/.rvm/gems/ruby-1.9.2-p0@rails3/gems/chef-0.9.8/lib/chef/knife.rb:127:in `run'
    from /home/matt/.rvm/gems/ruby-1.9.2-p0@rails3/gems/chef-0.9.8/lib/chef/application/knife.rb:118:in `run'
    from /home/matt/.rvm/gems/ruby-1.9.2-p0@rails3/gems/chef-0.9.8/bin/knife:25:in `<top (required)>'
    from /home/matt/.rvm/gems/ruby-1.9.2-p0@rails3/bin/knife:19:in `load'
    from /home/matt/.rvm/gems/ruby-1.9.2-p0@rails3/bin/knife:19:in `<main>'
```

PR #3638, opened 2015-07-24
[vsphere] Add support for cookie login

lwilke:
Hi,

this patch adds support for cookie auth to a vCenter server. To be exact i am only adding the RbVmomi option for this to the Fog::Vsphere credentials/options and changed the call to the connection constructor.

This is quite useful if you only want to login once and then work without a password, especially
the need to store a password somewhere in clear text is somewhat diminished.

Here is the link to the patch, that added this functionality to RbVmomi
https://github.com/rlane/rbvmomi/pull/22

cheers
    --lars

### DIFF
--- a/lib/fog/vsphere/compute.rb
+++ b/lib/fog/vsphere/compute.rb
@@ -7,7 +7,7 @@ module Fog
       requires :vsphere_username, :vsphere_password, :vsphere_server
       recognizes :vsphere_port, :vsphere_path, :vsphere_ns
       recognizes :vsphere_rev, :vsphere_ssl, :vsphere_expected_pubkey_hash
-      recognizes :vsphere_debug
+      recognizes :vsphere_debug, :vsphere_cookie
 
       model_path 'fog/vsphere/models/compute'
       model :server
@@ -358,6 +358,7 @@ module Fog
           require 'rbvmomi'
           @vsphere_username = options[:vsphere_username]
           @vsphere_password = 'REDACTED'
+          @vsphere_cookie   = options[:vpshere_cookie] || nil
           @vsphere_server   = options[:vsphere_server]
           @vsphere_expected_pubkey_hash = options[:vsphere_expected_pubkey_hash]
           @vsphere_is_vcenter = true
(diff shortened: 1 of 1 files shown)


geemus:
@nirvdrum could you take a look when you have a moment? Thanks!

nirvdrum:
I'm not opposed to the change, but a cookie seems just like another password . . . anyone that has it can authenticate.  That aside, it looks like this hasn't been merged into rbvmomi yet.  Is that accurate?

lwilke:
Hi,

* Kevin Menard wrote:
>    I'm not opposed to the change, but a cookie seems just like another
>    password . . . anyone that has it can authenticate. That aside, it looks

Yes and no, if your vCenter is doing auth against a central LDAP database, which
is used for various login purposes, things look different.

Besides the cookie or better token expires IIRC after 30 minutes.
So is much less dangerous to pass around.

Also in my case, with the token i only need to login once and both the Perl SDK
and fog work fine with it.

I am right now using fog with my change and stock vmware/rbvmomi 1.8.2.

>    like this hasn't been merged into rbvmomi yet. Is that accurate?

Ah sorry, the change was introduced into vmware/rbvmomi not the older rlane
version.

http://www.rubydoc.info/github/vmware/rbvmomi/RbVmomi/VIM
https://github.com/vmware/rbvmomi/commit/a1be0e6ea87bae7acca20a2e4293b9fb745cd394
https://github.com/vmware/rbvmomi/commit/7ecdadd107dfe608866d5f2974fcb14511bce41c

commit a1be0e6ea87bae7acca20a2e4293b9fb745cd394
Merge: d06b95b 7ecdadd
Author: cdickmann <christian.dickmann@gmail.com>
Date:   Sat Aug 10 19:10:53 2013 -0700

    Merge pull request #22 from subwindow/add_cookie

    Add support for connecting with a session cookie


### Here is some sample code

#!/usr/bin/env ruby
#
#require 'fog'
require '/home/me/repo/upstream/fog/lib/fog'
require 'pp'

credentials = {
  :provider                        => "vsphere",
  :vsphere_server                  => "vcenter",
  :vsphere_username                => ENV['USER'],
  :vsphere_password                => "",
  :vsphere_cookie                  => ENV['RBVMOMI_COOKIE'],
  :vsphere_expected_pubkey_hash    => "abcd",
  :vsphere_rev                     => "5.5",
}

connection = Fog::Compute.new(credentials)
pp connection.get_virtual_machine('HOST')

###

This line is a bit dirty

   :vsphere_password               => "",

but i need to satisfy the requires in lib/fog/vsphere/compute.rb
I did not know a way to say: require that password or cookie is set.

cheers
   --lars

nirvdrum:
Gotcha.  Thanks for the clarification on both fronts.

plribeiro3000:
@lwilke Would you be able to open this against `fog-vsphere` instead?

Thanks!

PR #3642, opened 2015-07-30
[openstack] move common things among services into core

d063130:
I noticed, that only compute and network use the core method `initialize_identity`, which enabled this services to be used with a v3 keystone - the other did not work.

Additionally I added the v3 parameters to the recognized parameters for all services and
instead of duplicating a lot of code, I took the chance to reuse the common stuff from core.


d063130:
sorry need to cherry-pick and reopen

PR #3645, opened 2015-08-02
[vsphere] Bug in full_path cluster attribute

slivik:
Create VM failed when there was a folder between Datacenter and Clusters. This bug seemed to be a typo only.

### DIFF
--- a/lib/fog/vsphere/requests/compute/list_clusters.rb
+++ b/lib/fog/vsphere/requests/compute/list_clusters.rb
@@ -41,7 +41,7 @@ module Fog
 
         def cluster_path(cluster, datacenter_name)
           datacenter = find_raw_datacenter(datacenter_name)
-          cluster.pretty_path.gsub(/(#{datacenter.name}|#{datacenter.hostFolder.name})\//,'')
+          cluster.pretty_path.gsub(/(#{datacenter_name}|#{datacenter.hostFolder.name})\//,'')
         end
       end
 


lib/fog/vsphere/requests/compute/list_clusters.rb
```diff
@@ -41,7 +41,7 @@ def cluster_attributes cluster, datacenter_name
 
         def cluster_path(cluster, datacenter_name)
           datacenter = find_raw_datacenter(datacenter_name)
-          cluster.pretty_path.gsub(/(#{datacenter.name}|#{datacenter.hostFolder.name})\//,'')
+          cluster.pretty_path.gsub(/(#{datacenter_name}|#{datacenter.hostFolder.name})\//,'')
         end
       end
```
plribeiro3000: If `datacenter`was loaded using the `datacenter_name` then its `name` value should be the same, no?

slivik: Yes, but in <code>datacenter_name</code> is full path to DC which has to be removed from full path to cluster and only relative path from DC to cluster must remain.

plribeiro3000: So you are saying that the object `datacenter`does return the full path but you only need the relative path which in this case is `datacenter_name`?

slivik: I was maybe not so understandable. Let me be more clear. This is full path to one of our clusters: <code>vCenter/Germany[folder]/GIO[datacenter]/host[vmware-obj]/GIO[folder]/MGMT[cluster]</code>.
The method <code>cluster_path</code> should return relative path to a cluster (<code>GIO[folder]/MGMT</code> - relative to datacenter) as it is awaited in <code>create_vm.rb</code> when discovering <code>resource_pool</code>. While "cluster.pretty_path" returns full path to cluster beginning with first folder under root, e.g. <code>Germany[folder]/GIO[datacenter]/host[vmware-obj]/GIO[folder]/MGMT[cluster]</code>, full path to datacenter needs to be subtracted. But <code>datacenter.name</code> returns only simple DC name (<code>GIO</code> in this case) while <code>datacenter_name</code> contains path to datacenter (<code>Germany/GIO[datacenter]</code>).

plribeiro3000: Got it. I guess those names are just not helping. `cluster_path` should be `cluster_relative_path` or something like that to be more explicit. Don´t need to comment on `pretty_path`. heheheh

Sorry to bother you but since i don´t have knowledge of vsphere i need some guidance. =)

This change seems pretty safe to me. Just to be sure, is there any place where `cluster_path` is used that might be expecting the behaviour you just mentioned is wrong?

slivik: I think not. IMHO this variable was introduced only to help set the full_path attribute of cluster (just because of the folder structure). You are right that it should be named a bit better.

plribeiro3000: Awesome. Thank you for the explanation and help!!

geemus: Thanks!

PR #3647, opened 2015-08-05
OpenStack compute add shelve related actions

Ladas:
Adding actions shelve, unshelve and shelve_off_load

Testing only shelve and unshelve, since shelve offload state
can be reached automaticallt dpending on settings in
shelved_offload_time in nova.conf.


TerryHowe:
lgtm

geemus:
Thanks!

PR #3648, opened 2015-08-05
[OpenStack] subnets vanilla options fix

d063130:
The selection of vanilla options in the subnet creation needs to be the same as in update.

Especially the `gateway_ip` is tricky (hence the special treatment beforehand), because if it is not present, then OpenStack chooses a default, if you set it to nil, it will be left empty (compare [apiref](http://developer.openstack.org/api-ref-networking-v2.html#createSubnet)). So it actually makes a difference there.

Additionally I added the `:allocation_pools`vanilla option to the subnet update, which was already present in create.

And along the way I removed some requires (probable from a copy&paste error) in the subnet model method `update`.

### DIFF
--- a/lib/fog/openstack/requests/network/create_subnet.rb
+++ b/lib/fog/openstack/requests/network/create_subnet.rb
@@ -38,9 +38,9 @@ module Fog
             'cidr'             => cidr,
             'ip_version'       => ip_version,
             'gateway_ip'       => options[:gateway_ip],
-            'allocation_pools' => options[:allocation_pools]  || [],
-            'dns_nameservers'  => options[:dns_nameservers]   || [],
-            'host_routes'      => options[:host_routes]       || [],
+            'allocation_pools' => options[:allocation_pools],
+            'dns_nameservers'  => options[:dns_nameservers],
+            'host_routes'      => options[:host_routes],
             'enable_dhcp'      => options[:enable_dhcp],
             'tenant_id'        => options[:tenant_id]
           }
--- a/lib/fog/openstack/requests/network/create_subnet.rb
+++ b/lib/fog/openstack/requests/network/create_subnet.rb
@@ -14,7 +14,7 @@ module Fog
           vanilla_options = [:name, :gateway_ip, :allocation_pools,
                              :dns_nameservers, :host_routes, :enable_dhcp,
                              :tenant_id]
-          vanilla_options.reject{ |o| options[o].nil? unless o == :gateway_ip }.each do |key|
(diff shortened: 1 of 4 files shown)


geemus:

TerryHowe:
:+1:

lib/fog/openstack/requests/network/create_subnet.rb
```diff
@@ -14,7 +14,7 @@ def create_subnet(network_id, cidr, ip_version, options = {})
           vanilla_options = [:name, :gateway_ip, :allocation_pools,
                              :dns_nameservers, :host_routes, :enable_dhcp,
                              :tenant_id]
-          vanilla_options.reject{ |o| options[o].nil? unless o == :gateway_ip }.each do |key|
+          vanilla_options.select{ |o| options.key?(o) }.each do |key|
```
Ladas: hm, might be good to extract this behavior to some helper, cause I think I saw this filtering on other places too. 

Should be always .key?, users should be free to send any value they need.

But that is not scope of this PR, maybe if you will be bored in the future, or somebody else can tackle it. :-)

lib/fog/openstack/requests/network/update_subnet.rb
```diff
@@ -24,11 +24,12 @@ class Mock
         def update_subnet(subnet_id, options = {})
           response = Excon::Response.new
           if subnet = list_subnets.body['subnets'].find { |_| _['id'] == subnet_id }
-            subnet['name']            = options[:name]
```
Ladas: hm, seems like a pretty bad way to update

since e.g. when you want to change a name, you will just send :name => 'something', so the check should be the same as in create (different fields, since some can't be updated)

But again, not your fault, it's not scope of this PR.

Ladas:
lgtm :+1: , comments I have are not really related to changes in this PR

lib/fog/openstack/requests/network/create_subnet.rb
d063130: Good advice, I saw that in other places, too - I will keep it in mind

lib/fog/openstack/requests/network/update_subnet.rb
d063130: It's the Mock, so I didn't really care much :)

Ladas: Ah, stupid me. :-) I overlooked the start of the mock class. :-)

geemus:
Thanks!

PR #3649, opened 2015-08-05
Fix comment: 5mins = 300s

ajlanghorn:
Fixes the comment to accurately reflect how many seconds are in five
minutes.

### DIFF
--- a/lib/fog/rackspace/examples/compute_v2/create_network.rb
+++ b/lib/fog/rackspace/examples/compute_v2/create_network.rb
@@ -58,7 +58,7 @@ puts "\nNow creating server '#{server.name}' connected the the Internet and '#{n
 
 begin
   # Check every 5 seconds to see if server is in the active state (ready?).
-  # If the server has not been built in 5 minutes (600 seconds) an exception will be raised.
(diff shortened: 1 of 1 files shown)


TerryHowe:
:+1:

plribeiro3000:
Hmm. actually the code is using `600`. Maybe it should say 10 minutes?

I don't remember exactly how the method `wait_for` syntax.

ajlanghorn:
@plribeiro3000 Oh, darn; you're right. I can either up the limit, or change the comment... which is preferred?

plribeiro3000:
I think its better to just change the comment. Since the code is this way it might have a reason. =)

Just looked at the implementation [here](https://github.com/fog/fog-core/blob/d117cd252d28b6b6d411043b6766413e260898db/lib/fog/core/wait_for.rb#L2) and it turns out that we have an interval of `600` and an interval of `5`. =)

ajlanghorn:
@plribeiro3000 Good call; thank you. I've repushed.

plribeiro3000:
Awesome. Thanks. 

I gotta have to take a look at this build. Sometimes it fails, sometimes not. =s

geemus:
Thanks!

PR #3650, opened 2015-08-06
[openstack] move common things among services into core

d063130:
cleanup of https://github.com/fog/fog/pull/3642

"I noticed, that only compute and network use the core method `initialize_identity`, which enabled this services to be used with a v3 keystone - the other did not work.

Additionally I added the v3 parameters to the recognized parameters for all services and
instead of duplicating a lot of code, I took the chance to reuse the common stuff from core."

### DIFF
--- a/lib/fog/openstack.rb
+++ b/lib/fog/openstack.rb
@@ -1,5 +1,6 @@
 require 'fog/openstack/compute'
 require 'fog/openstack/identity_v2'
+require 'fog/openstack/identity_v3'
 require 'fog/openstack/image'
 require 'fog/openstack/metering'
 require 'fog/openstack/network'
--- a/lib/fog/openstack/baremetal.rb
+++ b/lib/fog/openstack/baremetal.rb
@@ -6,10 +6,15 @@ module Fog
       SUPPORTED_VERSIONS = /(.)*/
 
       requires :openstack_auth_url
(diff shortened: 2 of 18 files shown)


TerryHowe:
I didn't go over it carefully, but definitely nice direction.

geemus:
Cool.

@Ladas could you take a peek as well? Thanks!

Ladas:
@d063130 hmm, something is wrong

OpenStack[:orchestration].stacks doesn't work

and 
OpenStack[:orchestration] claims it's not there, but when I first call OpenStack[:compute], then the orchestration is there, but .stacks still doesn't work. So there is some evil sharing of some variable somewhere inside. :-)

Could you check it out? I'll give it another test once you will figure it out. Thanks.

d063130:
Sorry, I had to force update the branch.

Thanks to @Ladas for the thorough look, I figured out what was wrong with the orchestration: missing collection requires and an unusual path handling, that is now aligned with the other services.

geemus:
Great!

@TerryHowe @Ladas mind giving in another once over post-fixes?

Ladas:
Looks good now. :+1: 

I also tested it works with keystone v3, or in other words keystone v3 doesn't work without this, cause in most services it filters out the domain param.

d063130 Good job with this fix, thank you for the patch.

geemus:
Thanks!

PR #3653, opened 2015-08-11
vCloud Director queued tasks should not count as "non_running".

adamleff:
In vCloud Air, created tasks end up in an initial state of "queued."
Fog currently treats these tasks as non_running and therefore stops
waiting for them.  Because the status is not yet successful, an
exception is raised.

I propose this change to treat a "queued" task as equivalent to
"running."


geemus:
Thanks!

PR #3655, opened 2015-08-12
Add missing require of identitty v3

Ladas:
Otherwise we get exception with missing
Fog::Identity::OpenStack::V3

### DIFF
--- a/lib/fog/openstack.rb
+++ b/lib/fog/openstack.rb
@@ -1,5 +1,6 @@
 require 'fog/openstack/compute'
 require 'fog/openstack/identity_v2'
+require 'fog/openstack/identity_v3'
 require 'fog/openstack/image'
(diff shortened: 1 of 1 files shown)


Ladas:
@geemus heya, can we get this one in before release? otherwise the keystone v3 fails on call

Fog::Identity::OpenStack::V3.new

dhague:
Good catch - thanks

TerryHowe:
:+1:

tnaoto:
:+1: 

I thought trying to I also just fix.

Ladas:
btw. I am experiencing also other issues.

e.g. console /bin/fog is not working with v3

and a weird thing that some services don't have endpoints, when I get token from v3

geemus:
Good catch, thanks!

PR #3658, opened 2015-08-13
Add boot_volume_size to RS compute_v2 Server Create

adecarolis:
Currently, fog hard-codes the size of Rackspace boot volumes to 100GB. 
This commit defines the boot_volume_size attribute (100GB is kept as a default).

### DIFF
--- a/lib/fog/rackspace/requests/compute_v2/create_server.rb
+++ b/lib/fog/rackspace/requests/compute_v2/create_server.rb
@@ -18,6 +18,7 @@ module Fog
         #   attachment of volumes to this server. Mutually exclusive with :volume_id or :volume_image_id. If provided, leave image_id
         #   as "". See http://developer.openstack.org/api-ref-compute-v2-ext.html#ext-os-block-device-mapping-v2-boot for details.
         # @option options [String] boot_volume_id Id of a pre-created bootable volume to use for this server. If provided, leave image_id as "".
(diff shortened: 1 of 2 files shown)


lib/fog/rackspace/requests/compute_v2/create_server.rb
```diff
@@ -96,12 +97,16 @@ def create_server(name, image_id, flavor_id, min_count, max_count, options = {})
                 Fog::Logger.warning(":boot_volume_id overrides :boot_image_id!")
               end
 
+              if options[:boot_volume_size]
```
geemus: I imagine this was probably here for debugging purposes, we should probably remove it before bringing this in?

adecarolis: I saw that similar warning for boot_volume_id therefore I thought the same was needed for boot_volume_size...

adecarolis: Would you like me to remove it?

geemus: Good question. @mdarby what do you think? I can see this being useful, but I don't want to end up with the command being too noisy. I guess at logger level you could filter these easily enough though.

mdarby:
:+1:

mdarby:
@geemus sorry for the delay -- could we get this merged today please?

geemus:
Sorry, honeymoon, merging now.

PR #3659, opened 2015-08-14
Openstack missing passing of unscoped_token

Ladas:
In the https://github.com/fog/fog/commit/803d46791f5cfe75c99084c49519af6d63248f9b
refactoring, the unscoped token passing to credentials has
been deleted. So adding unscoped token to 2 missing places.

Plus aligning the =>, so the code is more readable

### DIFF
--- a/lib/fog/openstack/core.rb
+++ b/lib/fog/openstack/core.rb
@@ -97,13 +97,15 @@ module Fog
       end
 
       def credentials
-        options =  { :provider => 'openstack',
-          :openstack_auth_url       => @openstack_auth_uri.to_s,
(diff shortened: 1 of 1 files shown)


TerryHowe:
Looks good, but makes me wonder why an unscoped token is treated any different than a regular auth token.

Ladas:
@TerryHowe they have probably removed unscoped token in keystone v3 (probably, because some say it should still work as in v2, but is not working like that now), but keystone v2 workflow should be something like

http://bodenr.blogspot.cz/2014/03/openstack-keystone-workflow-token.html

which is very weird :-)

geemus:
Thanks!

PR #3660, opened 2015-08-14
[Brightbox] Remove provider specific tests

tokengeek:
The `shindo` tests for Brightbox have been moved to the `fog-brightbox`
gem. They were initially left here to ensure `fog` was backwards
compatible when switching to `fog-brightbox` when we started splitting
out providers.


plribeiro3000:
:+1:

geemus:
Thanks, will be nice to slim this down further.

PR #3665, opened 2015-08-18
Rackspace temporary URL special options

duhast:
Hi!

Rackspace CloudFiles has some [custom options](http://docs.rackspace.com/files/api/v1/cf-devguide/content/TempURL_File_Name_Overrides-d1e213.html) for a private file's temporary URL. This update allows using both `inline` and `filename` options.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -81,6 +81,9 @@ namespace :test do
   task :openstack do
       sh("export FOG_MOCK=#{mock} && bundle exec shindont tests/openstack")
   end
+  task :rackspace do
+      sh("export FOG_MOCK=#{mock} && bundle exec shindont tests/rackspace")
+  end
   task :cloudstack do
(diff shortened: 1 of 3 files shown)


geemus:
@mdarby could you review? Thanks!

mdarby:
:+1:

geemus:
We are (at least for the time being) still trying to support 1.8.7 and it looks like this might have some syntax incompatibilities (appears to be colons vs hash rockets in a hash). Could you fix those? See: https://travis-ci.org/fog/fog/jobs/76061645#L3589

duhast:
@geemus done!

geemus:
Thanks!

duhast:
@geemus When do you plan to release a new `fog` version?

geemus:
We usually try to do at least once a month (the first Thursday of the month is the target), otherwise we occasionally will for critical stuff.

PR #3666, opened 2015-08-18
[WIP/RFC] initial support for OpenStack Volume API v2

majewsky:
Since the backwards-incompatible changes between v1 and v2 are tiny (only two attributes of volumes and snapshots have been renamed), I try to avoid code duplication and handle the changed attribute names gracefully.

This PR is in no way finished (there is no test coverage for v2, and I might also have to improve the get_service logic a bit), but I'm posting it already for some early feedback. I'll post line comments on the exact parts.

### DIFF
--- a/lib/fog/openstack/models/volume/volume.rb
+++ b/lib/fog/openstack/models/volume/volume.rb
@@ -6,8 +6,10 @@ module Fog
       class Volume < Fog::OpenStack::Model
         identity :id
 
-        attribute :display_name,        :aliases => 'displayName'
-        attribute :display_description, :aliases => 'displayDescription'
+        # NOTE: The attributes "name" and "description" are called
+        # "display_name" and "display_description" in API version v1.
+        attribute :name,                :aliases => ['display_name', 'displayName', :display_name]
+        attribute :description,         :aliases => ['display_description', 'displayDescription', :display_description]
         attribute :metadata
         attribute :status
         attribute :size
@@ -21,8 +23,8 @@ module Fog
         attribute :tenant_id,           :aliases => 'os-vol-tenant-attr:tenant_id'
 
         def save
-          requires :display_name, :size
-          data = service.create_volume(display_name, display_description, size, attributes)
+          requires :name, :size
+          data = service.create_volume(name, description, size, attributes)
           merge_attributes(data.body['volume'])
           true
         end
@@ -41,6 +43,13 @@ module Fog
 
         def ready?
           status == 'available'
+        end
+
+        def display_name
+          name
+        end
+        def display_description
+          description
         end
       end
     end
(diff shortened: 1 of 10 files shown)


lib/fog/openstack/models/volume/volume.rb
```diff
@@ -6,8 +6,10 @@ class OpenStack
       class Volume < Fog::OpenStack::Model
         identity :id
 
-        attribute :display_name,        :aliases => 'displayName'
-        attribute :display_description, :aliases => 'displayDescription'
+        # NOTE: The attributes "name" and "description" are called
+        # "display_name" and "display_description" in API version v1.
+        attribute :name,                :aliases => ['display_name', 'displayName', :display_name]
+        attribute :description,         :aliases => ['display_description', 'displayDescription', :display_description]
```
majewsky: `:aliases` are commonly given as strings, but in here, I also needed to give it as a symbol. If I don't, loading of objects via e.g. `@service.volumes.all(...)` fails when calling `object.merge_attributes({ 'display_name' => 'foo', ... })`. I'm not sure if this is a problem with fog-core (should either strings or symbols be normalized at some point inside `Fog::Attributes`?) or if I'm just the first to handle a renaming of attribute names in an API? (If there is a way to do this right, pointers are highly appreciated.)

lib/fog/openstack/models/volume/volume.rb
```diff
@@ -42,6 +44,13 @@ def extend(size)
         def ready?
           status == 'available'
         end
+
+        def display_name
+          name
+        end
+        def display_description
+          description
+        end
```
majewsky: This is also obviously ugly. The `spec/fog/openstack/volume_spec.rb` uses these accessors, but giving `display_name` and `display_description` as aliases for `name` and `description` above does not seem to generate them. Again, is there a nicer way to handle this?

geemus:

TerryHowe:
I haven't tried it and I don't know about the about the accessors or aliases issues, but from an OpenStack v1/v2 perspective, lgtm.

lib/fog/openstack/models/volume/volume.rb
Ladas: not sure how aliases works

you can write this shorter with:
alias_method :display_name, :name

lib/fog/openstack/models/volume/volume.rb
Ladas: might be cleaner to create v1 and v2 models, with extracting the same behavior to base class

Ladas:
@majewsky looks good so far, we might need v2 namespace, since I see that there are a few v2 specific API calls. If v2 is extending a overriding v1 (so v1 is a subset) we could model it like that, with a little code duplication.

majewsky:
Just spoke to @dhague; he will look into making this design a bit cleaner and easier to extend for v2-only models/requests. I'll therefore close this one.

Ladas:
cool, I saw he did also glance, awesome! thank you guys

PR #3669, opened 2015-08-21
Adding #get_by_id for Fog::Compute::VcloudDirector::VmCustomizations

adamleff:
This method is necessary for functions such as #reload to function on
a VM customization.  This also allows users who set and save a customization
to re-fetch it to get values such as the VM administrator password.


geemus:
Thanks!

PR #3670, opened 2015-08-22
[openstack]reset api path by network model set tenant 

pyama86:
Api path has been reset called the network set tenants

call authenticateby network set tenant 
https://github.com/fog/fog/blob/master/lib/fog/openstack/requests/network/set_tenant.rb#L8

reset api path by openstack core
https://github.com/fog/fog/blob/master/lib/fog/openstack/core.rb#L151

api path must be set 
https://github.com/fog/fog/blob/master/lib/fog/openstack/network.rb#L242

### DIFF
--- a/lib/fog/openstack/network.rb
+++ b/lib/fog/openstack/network.rb
@@ -238,14 +238,7 @@ module Fog
           @connection_options     = options[:connection_options] || {}
 
           authenticate
-
-          @path.sub!(/\/$/, '')
-          unless @path.match(SUPPORTED_VERSIONS)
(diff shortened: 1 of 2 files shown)


geemus:

TerryHowe:
lgtm, just not happy in general how versions are handled in Fog/openstack but that is for another day.  I just wish it was smarter about selecting the right version automatically based on the service catalog, the versions API and user preferences.

Ladas:
lgtm :+1:

geemus:
Thanks!

@TerryHowe yeah, noted. Definitely open to discussing more.

PR #3671, opened 2015-08-24
Cloud Director supports multiple peer subnets.

joshmyers:
As with local_subnets, Vcloud-director supports
multiple peer_subnets. This is a needed feature if
you want to have multiple subnets talk
over the same VPN tunnel.


geemus:
Thanks1

PR #3674, opened 2015-08-26
Fix attach router to also use port id if passed

dhague:


TerryHowe:
Alternatively, it could just take subnet id with optional port id, but this should work.

geemus:
Is subnet_id required either way? Might be a less major change to just merge in the options?

dhague:
I believe that passing subnet_id alone as a string (as opposed to within a hash) is actually a bug, but the behaviour was retained here for backward compatibility in case anyone is already using it.

geemus:
Fair enough. Thanks!

PR #3677, opened 2015-08-27
[OpenStack] Enable list_snapshots_detailed request

blomquisg:
Deprecation warnings suggest:

> [fog][DEPRECATION] Calling `OpenStack[:volume].list_snapshots(true)` is deprecated, use `.list_snapshots_detailed` instead

However, `Fog::OpenStack::Volume` does not respond to `.list_snapshots_detailed`.
Adding that to the list of requests available to the OpenStack Volume service.

Added a super-simple test to make sure the Volume service now at least responds to `list_snapshots_detailed`.

### DIFF
--- a/lib/fog/openstack/volume.rb
+++ b/lib/fog/openstack/volume.rb
@@ -45,6 +45,7 @@ module Fog
 
       request :create_volume_snapshot
       request :list_snapshots
+      request :list_snapshots_detailed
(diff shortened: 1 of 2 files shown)


geemus:
Thanks!

blomquisg:
Thanks @geemus!

PR #3678, opened 2015-08-27
Fix deprecation warnings

blomquisg:
A couple fixes to internal calls that were causing deprecation warnings.

The warnings I was seeing were:

> Calling OpenStack[:compute].list_security_groups(server_id) is deprecated, use .list_security_groups(:server_id => value) instead

And,

> Calling OpenStack[:orchestration].list_resources(stack, options) is deprecated, call .list_resources(:stack => stack) or .list_resources(:stack_name => value, :stack_id => value) instead

### DIFF
--- a/lib/fog/openstack/models/compute/server.rb
+++ b/lib/fog/openstack/models/compute/server.rb
@@ -212,7 +212,7 @@ module Fog
         def security_groups
           requires :id
 
-          groups = service.list_security_groups(id).body['security_groups']
+          groups = service.list_security_groups(:server_id => id).body['security_groups']
 
           groups.map do |group|
             Fog::Compute::OpenStack::SecurityGroup.new group.merge({:service => service})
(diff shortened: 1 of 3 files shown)


geemus:
Looks like there may be some test failures here, could you check into the travis run above? Thanks!

blomquisg:
Yeah, I'll see if I can reproduce these locally.

geemus:
Thanks, just let me know if you have questions. A quick glance at errors didn't make it totally obvious to me how it was related, but hopefully it will be clear with your closer inspection.

blomquisg:
Found it ... the `Fog::Compute::OpenStack::Mock#list_security_groups` needed to be updated to handle the newer method signature.

geemus:
Thanks!

blomquisg:
Thanks @geemus!

PR #3682, opened 2015-09-03
[openstack] Snapshot manipulation using Openstack Cinder service

tunglam14:
In, current version, snapshot manipulation is using Nova Endpoint, it couldn't show metadata of a snapshot, but Cinder Endpoint can.

So, I've cloned Snapshot method from Compute model to Volume model for using Cinder Endpoint.

The result when get Snapshot by Compute model:

```
=>   <Fog::Compute::OpenStack::Snapshot
    id="91361632-5ea8-4ca5-8e35-b8dd4568d4f8",
    name="1231231231_rootdisk@2015-08-04",
    description="",
    volume_id="c63ff7aa-4858-41ea-8bac-b09a9c0a155f",
    status="available",
    size=20,
    created_at="2015-08-04T04:08:09.000000"
  >
```

^ No metadata here

And, the result from Volume model:

```
<Fog::Volume::OpenStack::Snapshot
    id="91361632-5ea8-4ca5-8e35-b8dd4568d4f8",
    display_name="1231231231_rootdisk@2015-08-04",
    display_description="",
    volume_id="c63ff7aa-4858-41ea-8bac-b09a9c0a155f",
    status="available",
    size=20,
    created_at="2015-08-04T04:08:09.000000",
    metadata={"type"=>"backup", "job_id"=>"55bee7bcf4ebf4920800000e"}
  >
```

### DIFF
--- a/lib/fog/openstack/models/volume/snapshot.rb
+++ b/lib/fog/openstack/models/volume/snapshot.rb
@@ -1,0 +1,32 @@
+require 'fog/openstack/models/model'
+
+module Fog
+  module Volume
+    class OpenStack
+      class Snapshot < Fog::OpenStack::Model
+        identity :id
+
+        attribute :display_name
+        attribute :display_description
+        attribute :volume_id
+        attribute :status
+        attribute :size
+        attribute :created_at
+        attribute :metadata
+
+        def save(force=false)
+          requires :volume_id, :name, :description
+          data = service.create_volume_snapshot(volume_id, name, description, force)
+          merge_attributes(data.body['snapshot'])
+          true
+        end
+
+        def destroy
+          requires :id
+          service.delete_snapshot(id)
+          true
+        end
+      end
+    end
+  end
+end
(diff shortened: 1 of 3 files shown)


geemus:
@Ladas @TerryHowe review please? Thanks!

lib/fog/openstack/models/volume/snapshot.rb
```diff
@@ -0,0 +1,32 @@
+require 'fog/openstack/models/model'
+
+module Fog
+  module Volume
+    class OpenStack
+      class Snapshot < Fog::OpenStack::Model
+        identity :id
+
+        attribute :display_name
+        attribute :display_description
```
TerryHowe: v1 cinder is already deprecated and v2 uses name and description over display_name and display_description.  It'd be nice if these two at least were aliased.

TerryHowe:
I don't completely understand the overall direction for version compatibility across models.  I'm kind of in favor of not trying to hide versions to the user, but not sure how that should be managed with Fog.

Ladas:
Late in the game. I've totally missed this due to my vacation. :-)

@tunglam14 If you decide to reopen, you could solve the cinder v1/v2 the same as in this PR https://github.com/fog/fog/pull/3666

PR #3684, opened 2015-09-04
[#3099] Remove Nokogiri constraint

tokengeek:
Since 1.8.7 compatibility has been dropped, we no longer need to lock to
the last usable version. This should allow usage of later Rails which
require newer nokogiri versions.


geemus:
Thanks!

PR #3686, opened 2015-09-08
Openstack volumes adding volume type management

Ladas:
Opensopestack_volumes_manage_volume_typetack volumes adding volume type management


geemus:
Thanks!

PR #3689, opened 2015-09-14
Fix distributed port group switches support for cloning with multiple networks

iNecas:
We were not passing the datacenter parameter when searching for networks,
which caused issues when searching for distributed switches attributes

Also, I've extracted the virtualswitch parameter to the network model,
which allows to specify the virtualswitch value for the network interface
based on the network data later. Our workflow is letting the user
to choose what network should be the nic attached to and then creating
the networks correspondingly and we use the network object for that.

### DIFF
--- a/lib/fog/vsphere/requests/compute/list_networks.rb
+++ b/lib/fog/vsphere/requests/compute/list_networks.rb
@@ -20,10 +20,11 @@ module Fog
 
         def network_attributes network, datacenter
           {
-            :id         => managed_obj_id(network),
-            :name       => network.name,
-            :accessible => network.summary.accessible,
-            :datacenter => datacenter,
+            :id            => managed_obj_id(network),
+            :name          => network.name,
+            :accessible    => network.summary.accessible,
+            :datacenter    => datacenter,
+            :virtualswitch => network.class.to_s == "DistributedVirtualPortgroup" ? network.config.distributedVirtualSwitch.name : nil
           }
         end
       end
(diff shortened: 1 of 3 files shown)


geemus:
@nirvdrum could you take a peak when you have a moment? Thanks!

eLobato:
@nirvdrum This is holding up a feature in Foreman, https://github.com/theforeman/foreman/issues/2714, would you mind taking a look? Thanks!

nirvdrum:
Sorry, I missed the first ping. Just to set expectations, I probably won't be able to look at this until the weekend.

lib/fog/vsphere/requests/compute/list_networks.rb
```diff
@@ -20,10 +20,11 @@ def raw_networks(datacenter_name)
...
           {
-            :id         => managed_obj_id(network),
-            :name       => network.name,
-            :accessible => network.summary.accessible,
-            :datacenter => datacenter,
+            :id            => managed_obj_id(network),
+            :name          => network.name,
+            :accessible    => network.summary.accessible,
+            :datacenter    => datacenter,
+            :virtualswitch => network.class.to_s == "DistributedVirtualPortgroup" ? network.config.distributedVirtualSwitch.name : nil
```
nirvdrum: We can fix this after, but calling `class.to_s` seems like an odd pattern to me. `class.name` is the typical way I've seen things done like this.

dLobatog:
@nirvdrum Thanks!

PR #3695, opened 2015-09-17
URL encoding fix for vCloud Director API

surminus:
This change does two things:

 - Changes the options passed into the request in `get_execute_query.rb` to a hash rather an array
 - Bugfixes an issue where our version of the vCloud Director API is not able to correctly parse an option we are requesting. More detail is within the commit.


geemus:
Thanks!

PR #3697, opened 2015-09-22
[openstack] Glance v2 API

dhague:
This is implemented in a similar style to Identity V3 - the core Fog::Image::OpenStack class is now just a wrapper which returns either a ImageV1 or ImageV2, choosing the latest version according to the OpenStack service catalog returned at authentication time.


geemus:
Thanks!

lib/fog/openstack/core.rb
```diff
@@ -572,6 +572,33 @@ def self.get_supported_version(supported_versions, uri, auth_token, connection_o
       version
     end
 
+    def self.get_supported_version_path(supported_versions, uri, auth_token, connection_options = {})
```
Ladas: this is awesome, could you put this code also for keystone v3?

I've hit an issue that keystone endpoint-list doesn't have to contain url for right version, so it's safer to discover it.

PR #3698, opened 2015-09-23
support DigitalOcean v2 compute API

bryanl:
This commit adds support for the DigitalOcean v2 API. It implements all the compute actions.


grossjo:
Thanks, I see the tests are failing.
I've started working on fixing some of them and adding to them.
I can start on my own fork, or if you want to give me access to your's I can commit directly to your branch.

bryanl:
@grossjo, I added you as a collaborator.

grossjo:
I have fixed all the existing tests and problems with the compute classes.
I'm going to start adding more tests.

athirn:
What's the next step to getting this approved?  Pretty important for Fog to support the DigitalOcean v2 API.  thanks

bryanl:
@athirn, @grossjo was adding tests. Then I guess it's up for an official review. Not sure who will do that.

grossjo:
I have commit rights to the fog project, so once the tests are finished, i'll give this a +1.
I've only been writing tests and fixing any anomalies that they find find, but i'm sure we can get one more person from fog to review it as well.

seanhandley:
I've given it a 20 minute read-through and it all looks sane and clean to me :+1: 

I have one question, though it's more about intended behaviour than code quality:

The server model exposes a public IP address attribute, but is only able to return an IPv4 address. Since Digital Ocean supports v6 (and we even have a server action to enable v6 addressing), wouldn't it make sense to return the v6 address if IPv6 has been enabled on the droplet?

grossjo:
We should probably expose explicit ipv6 and ipv4 public ip methods. I think the current implementation would be easiest for people to transition, base on the v1 code and other behaviour on projects in fog.

I'm not sure what fog's ideas are for  ipv6 support and how to expose this throughout the clouds it supports.

geemus:
Explicit methods sound good. We should probably not change the existing behavior though. So perhaps leaving that as is, but adding an alias to that for ipv4 to be more explicit (and then adding ipv6 methods alongside where available). I guess it might even be nice to add ipv6 methods where it isn't available to provide more concrete/explicit feedback (ie it just raises ipv6-unsupported or something).

bryanl:
@geemus and @grossjo, so what's the consensus. How should it be changed or augmented?

geemus:
My inclination would be:

1. Leave current stuff as-is
2. Add ipv4 related methods, which probably just call to existing stuff
3. Add ipv6 related methods that do the new thing
4. (someday/maybe) change the default and/or remove ipv4 stuff

seanhandley:
Given that the v1 API is due to be turned off in 6 weeks, I'm personally happy to see this merged now on the assumption that additional test coverage will be added later.

Even if that means a couple of potential issues; it's better to have support that mostly works, than no support at all.

Unless you guys have any reservations @grossjo @geemus, I think this should get merged.

grossjo:
I feel the same way. @geemus what are your thoughts?

geemus:
Sure, I think I agree. My concern just being that, if tests don't happen now, perhaps they won't happen at all (which would be unfortunate). Leaving it out in the mean time seems bad given current timeline though.

geemus:
@bryanl just missed you a couple times at Strange Loop. Hope things are going well.

geemus:
FYI due to ongoing work to drop 1.8.x and probably at least early 1.9.x series rubies, this will probably hang out in an RC for a bit before going to production. I will probably shoot for cutting a full-on-release in 2-3 weeks after I'm back from my honeymoon.

seanhandley:
Congrats @geemus - have a great time :wedding: :palm_tree: :sunny:

geemus:
Thanks!

PR #3701, opened 2015-09-25
fix linode kernel selection bug

b0ric:
I'm using knife-linode which depends on fog.
And it turned out that it doesn't pay attention to kernel ID provided. It instead just selects the first kernel from kernel list response.
This pull request is to fix the issue.


geemus:
Thanks!

Issue #3702, opened 2015-09-25
[openstack] bug retrieving list server actions


tunglam14:
I can't list actions of a server through `server_actions` method

```
Fog::Compute.new(arg).server_actions(server_id)
```

It return 404 errror.

I've digged in to the code and found it's calling to `/v2/tenant_id/servers/server_id/actions`  [at here](https://github.com/fog/fog/blob/master/lib/fog/openstack/requests/compute/server_actions.rb#L15). 
It's wrong, right? Because by the [Openstack API Documentation](http://developer.openstack.org/api-ref-compute-v2-ext.html#listActions), the API url is `/v2/​{tenant_id}​/servers/​{server_id}​/os-instance-actions`

Can you please take a look and confirm this.

Thanks

geemus:
@TerryHowe @Ladas could you guys advise? Thanks!

Ladas:
@tunglam14 seems what you are saying is correct, according to doc. The 'actions' is probably some historical API call, I found OpenStack wiki mentioning that.

tunglam14:
@Ladas Thanks, as PR was merged, i'm going to close this issue.

PR #3703, opened 2015-09-25
[openstack] List server actions fix

tunglam14:
By issue #3702 

Fix list server action endpoint, following the doc http://developer.openstack.org/api-ref-compute-v2-ext.html#listActions
Update Mock data


geemus:
@TerryHowe @Ladas or actually, maybe review this instead? Thanks!

Ladas:
@geemus looks good to me :+1: 

There is a weird inconsistency, that was there before, that request is loading body. It's not usually done in request. Might be worth of future refactoring. Not scope of the PR though :-)

geemus:
Thanks!

PR #3706, opened 2015-09-28
Limit 1.9.3 tests to use older Net::SSH

tokengeek:
Net::SSH 3.0 dropped support for Ruby 1.9.3 so older Rubies should
declare the dependency to avoid bundling problems.


plribeiro3000:
:+1:

PR #3707, opened 2015-09-29
Remove support for Ruby 1.9.3

tokengeek:
We had hoped to support Ruby 1.9.3 a bit longer even past its official
EOL but with our dependencies starting to drop support for 1.9 we end up
with several problems.

* Either we lock down our dependencies and block applications from using
  newer gems which is bad. We tried this with `nokogiri` for 1.8 and
  that prevented updates to Rails 4.0
* We could support 1.9 via the dependencies in a version dependent
  `Gemfile` but for each dependency that breaks support we have to push
  out changes to every fog repo with the new dependency.

Given the age and status of Ruby 1.9, we have decided to drop support
for Ruby 1.9.3 now along with 1.8 and 1.9.2.

If you are on an application that requires 1.9.3 and `fog` then you
should lock your dependencies to the `1.x` gems.


plribeiro3000:
:+1:

geemus:
Thanks!

PR #3709, opened 2015-09-30
Require fog-google versions compatible w/ > 1.9.3

icco:


plribeiro3000:
I wonder if we need to be pessimistic here. Maybe we could use `>=` and avoid future need to update the dependencies. This way the user can tweak the provider versions for its own need. What do you think?

icco:
That's fair, I'm totally down with that. I also fucked up that release by not adding the ruby version limiter. Will fix.

icco:
Fixed @plribeiro3000.

plribeiro3000:
Awesome. Thx for bearing with me @icco

geemus:
Thanks!

PR #3713, opened 2015-10-02
[openstack] Make sure that Identity v3 authentication uses a v3 endpoint

dhague:
Also, some IDv3 test refactoring - extract strings to constants and allow OS_AUTH_SERVICE env var
to set the openstack_service_type option (e.g. to identityv3).
Re-recorded VCRs against a new DevStack.


dhague:
@geemus @TerryHowe @Ladas Any comments? Any reason not to merge?

Ladas:
@dhague I haven't try this code, but seems it should caught error I've seen sooner now. Which is when you have v3 auth but v2 path in endpoints. Currently it failed when talking to endpoint. So looks good to me :+1: 

For the near future, I would like to change this to discovery of endpoint URL based on auth url. Seems like the OpenStack community is slowly aiming for not specifying version in keystone endpoint-list. So similar what you did for Glance AFAIK.

dhague:
I agree on the "next generation" approach to doing this by calling the URL to get the version from the metadata. I'll add it to my backlog :)

Ladas:
@dhague if you can make it, that would be awesome :+1:  thank you

geemus:
Thanks!

PR #3716, opened 2015-10-05
Ensure root user behaves like Rackspace defaults

### LINKED ISSUE #3715: Don't lock root user by default
We shouldn't be locking the root user by default in the Rackspace compute drivers. I think this is pretty counterintuitive, and "for better security," as the original PR https://github.com/fog/fog/pull/2492 says, seems like a very poor argument. I think the default should be to leave the server as-is (which may or may not be a locked root user), so that tools work, and allow the user/caller to opt-in to this behavior.

The argument that users should opt-in to this behavior is especially strong given that this behavior is different than the default behavior of using Rackspace's control panel and different than calling the Rackspace APIs.

martinb3:
Swap the logic for `:no_passwd_lock` making it `:passwd_lock`. Per #3715, we really should make our defaults the same as Rackspace's. Because so many projects rely on the fog library, and can't all add this option to pass through, we should default to the same behaviors you'd get on vanilla Rackspace.

This change is backwards compatible, but still warns those who have explicitly set :no_passwd_lock to any specific value.

Fixes #3715. RE: #2494.

### DIFF
--- a/lib/fog/rackspace/models/compute_v2/server.rb
+++ b/lib/fog/rackspace/models/compute_v2/server.rb
@@ -619,7 +619,11 @@ module Fog
         end
 
         def password_lock
-          "passwd -l #{username}" unless attributes[:no_passwd_lock]
+          if !attributes[:no_passwd_lock].nil?
+            Fog::Logger.warning("Rackspace[:no_passwd_lock] is deprecated since it is now the default behavior, use Rackspace[:passwd_lock] instead")
+          end
+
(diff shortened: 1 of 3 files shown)


seeafish:
+1
This is currently not in line with how Rackspace does post-config does, and thus breaks builds initiated by any fog related tool, where a key is not present/available.

geemus:
@mdarby could you review? Thanks!

mdarby:
:+1:

geemus:
Thanks!

PR #3717, opened 2015-10-08
Using Relative paths

batlongthien:


tokengeek:
Not sure what this PR is covering. Suspect it's here by mistake.

We certainly can't be removing all the gemspec references for bundler
references as then the gem won't work.

plribeiro3000:
agree @tokengeek

PR #3720, opened 2015-10-13
change get_raw replacement regex from />/ to /=>/ for output XML

eucher:
OpenNebula template may include RAW data. And if RAW data is like this:
```xml
RAW=["DATA"=>"<cpu match='exact'><model fallback='allow'>core2duo</model></cpu>", "TYPE"=>"kvm"]
```
Then current implementation replaces all '>' with ''. So output template will be:
```xml
RAW=["DATA"="<cpu match='exact'<model fallback='allow'core2duo</model</cpu", "TYPE"="kvm"]
```
And it generates an error. So i patch it to generating output like this:
```xml
RAW=["DATA"="<cpu match='exact'><model fallback='allow'>core2duo</model></cpu>", "TYPE"="kvm"]
```


geemus:
Thanks!

Issue #3728, opened 2015-10-22
[vsphere] - unable to add more than 7 disks


fherbert:
Due to the SCSI controller taking id 7, when creating a vm with more than 7 disks, and if one (generally disk #8) of those disks are set to use unitNumber = 7, then creating the vm will fail.

A relatively easy fix can be done to lib/fog/vsphere/requests/compute/create_vm.rb. I'll submit a pull request shortly.

plribeiro3000:
I´m not that into `vsphere` so perhaps you could help me out here. Where does this SCSI controller come from?

fherbert:
Any disk attached to a vm requires a scsi controller, it is created when the vm is created, before the disks are created. Check this adaptec page for info on scsi ID's http://ask.adaptec.com/app/answers/detail/a_id/689/~/can-the-scsi-id-of-an-adaptec-scsi-controller-card-be-changed%3F
The current fog code creates a single scsi controller just before the disks are created, you don't get the option to set the controller ID.

plribeiro3000:
Awesome. Thx for the explanation.

geemus:
I believe this is closed by #3729

PR #3729, opened 2015-10-22
[vsphere] - unable to add more than 7 disks

fherbert:
Hi, This is a fix for creating a vm with more than 7 disks. The issue exists because the SCSI controller takes ID 7, and the fog library creates disks generally based on array key (in the case of Foreman anyway). #3728

### DIFF
--- a/lib/fog/vsphere/requests/compute/create_vm.rb
+++ b/lib/fog/vsphere/requests/compute/create_vm.rb
@@ -122,18 +122,23 @@ module Fog
         end
 
         def create_disk disk, index = 0, operation = :add, controller_key = 1000
+          if (index > 6) then
+            _index = index + 1
+          else
+            _index = index
+          end
           payload = {
             :operation     => operation,
             :fileOperation => operation == :add ? :create : :destroy,
             :device        => RbVmomi::VIM.VirtualDisk(
-              :key           => disk.key || index,
+              :key           => disk.key || _index,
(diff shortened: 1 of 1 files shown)


lib/fog/vsphere/requests/compute/create_vm.rb
```diff
@@ -122,18 +122,23 @@ def controller_get_shared_from_options options
         end
 
         def create_disk disk, index = 0, operation = :add, controller_key = 1000
+          if (index > 6) then
+            _index = index + 1
+          else
+            _index = index
+          end
```
plribeiro3000: Isn´t this the case to only increment if index equals 7?

Maybe something like:

```ruby
index += 1 if index == 7
```

would fit better. Make sense?

fherbert: @plribeiro3000  - Only incrementing the index for 7 will resulting in losing a drive (if you have ore than 7 drives)
What about the actual drive that has the index of 8, this will be ignored by only incrementing if index == 7, because the drive with initial index of 7 will now be the drive with index of 8.
Basically we need to create a 'hole' in the array at index 7.

plribeiro3000: Got it. Make sense. :+1:

plribeiro3000:
Thank you for your efforts and time bearing with me. :+1:

geemus:
Thanks!

Issue #3718, opened 2015-10-08
Linode API changes


RomanKapitonov:
I've been looking into Linode's API implementation. There are some changes to the API so some of the API calls are not up-to-date. `avail.kernels` is one of them. Here is the link to the api call https://www.linode.com/api/utility/avail.kernels. The implementation does not contain currently available parameters for the call: https://github.com/fog/fog/blob/master/lib/fog/linode/requests/compute/avail_kernels.rb

`kernelId` should be eliminated. `isXen` and `isKVM` should be passed as parameters to the API call. There may also be other changes but this is still to be figured out.

geemus:
Thanks for the heads up. We would certainly welcome any help in this area. I'm happy to provide guidance, but am not presently using linode so don't feel well placed to fix it myself.

philr:
I have created pull request #3733 to update the implementation of the `avail.kernels` API call. The pull request also updates the other "avail" API requests to match the currently documented behaviour.

I've not looked for changes and differences to any other Linode API calls.

plribeiro3000:
:+1: for the PR!

RomanKapitonov:
:+1:

geemus:
Cool, going to close this for now as we are not presently aware of other things that are missing (and we can always address them individually/later as they come up). Thanks!

PR #3733, opened 2015-10-26
Update Linode "avail" API methods

philr:
This pull request updates the wrappers for the Linode "avail" API methods to match the currently documented behaviour:

- Add an `abbr` attribute to the DataCenter model.
- Update the `avail_kernels` request and `Kernel` model to reflect the removed `kernelId` parameter, newly added `isXen` and `isKVM` parameters and additional returned attributes.
- Update the `avail_linodeplans` request and `Flavor` model to add new attributes and change the mocked data center ids to match those currently returned.
- Add support for the `avail.nodebalancers` API call.
- Fix the signature of the `avail_stackscripts` mock method so that it matches the real method.

This resolves the specific issue with `avail.kernels` not being up to date raised in #3718.

This also modifies the fix made in #3701. The underlying reason for the issue raised in #3701 is that Linode changed the `avail.kernels` API call so that it no longer accepts a `kernelId` parameter. Attempting to pass the parameter now causes all the available kernels to be returned instead of just the one that was requested. This pull request modifies the `avail_kernels` request and `Kernels` collection so that the `kernelId` parameter is no longer included in the request.


philr:
The Travis CI build failed, but this appears to be failures that were already occurring on master with `Fog::Image::OpenStack`.

lib/fog/linode/models/compute/kernels.rb
```diff
@@ -11,20 +11,27 @@ def all
           load kernels
         end
 
+        def all_kvm
+          load kernels(:isKVM => 1)
+        end
+
+        def all_xen
+          load kernels(:isXen => 1)
+        end
```
plribeiro3000: Not sure about those methods. Maybe we should have filters on the `all` method?

```ruby
def all(options = {})
  load kernels(options)
end
```

What do you think?

lib/fog/linode/requests/compute/avail_kernels.rb
```diff
@@ -5,27 +5,30 @@ class Real
...
         #   * body<~Array>:
         # TODO: docs
-        def avail_kernels(kernel_id=nil)
-          options = {}
-          if kernel_id
-            options.merge!(:kernelId => kernel_id)
-          end
+        def avail_kernels(options={})
+          # avail.kernels used to accept a kernelId parameter (now removed)
+          raise Fog::Errors::Error.new('avail_kernels no longer accepts a kernelId parameter') unless !options || options.is_a?(Hash)
```
plribeiro3000: One can't list all kernels without any filters?
Like `options = nil`?

lib/fog/linode/models/compute/kernels.rb
philr: @plribeiro3000 I wasn't sure about this either. I'm new to the fog project, so I'm not sure what the usual convention would be. I'm happy to change it per your suggestion if you think that would be better.

plribeiro3000: I was thinking more about backwards compatibility in the future. It would be easier to remove an option then a method. But thats kind of my own opinion here. Not exactly something to be followed.

lib/fog/linode/requests/compute/avail_kernels.rb
philr: @plribeiro3000 `options = nil` and `options = {}` are both allowed. Both will return all the available kernels. The exception is raised if `options` is not `nil` and is not a `Hash`. This is to catch cases where callers may be using the old `kernelId` method signature.

plribeiro3000: Yep. Thats right. I always get confused with `unless !`. :grimacing: 
Maybe you could use a idiom like `if options.is_a?(Integer)`?

lib/fog/linode/models/compute/kernels.rb
philr: @plribeiro3000 using `options` on a single `all` method would also be better if you wanted to dynamically select either KVM or Xen kernels. I have committed e73242e32a68008bcecaa669df23cb6af266ccb8 to remove `all_kvm` and `all_xen` and add an `options` parameter to `all`.

lib/fog/linode/requests/compute/avail_kernels.rb
philr: @plribeiro3000 It is possible that any existing uses of `avail_kernels` could be using types other than `Integer` for the `kernelId` (Excon calls `to_s` on parameter values). I therefore prefer to reject anything that is not a `Hash` or `nil`.

lib/fog/linode/models/compute/kernels.rb
plribeiro3000: Nice.

lib/fog/linode/requests/compute/avail_kernels.rb
plribeiro3000: Sure. =)

geemus:
Thanks!

PR #3738, opened 2015-10-28
[openstack] Basic caching support for auth tokens, domains & projects

dhague:
Fairly conservative & secure for now - 30 second TTL on cache entries, and the auth token is part of the cache key.


dhague:
Oops - looks like I need to re-record the VCRs

geemus:
Thanks!

lib/fog/openstack/core.rb
```diff
@@ -504,17 +503,19 @@ def self.retrieve_tokens_v3(options, connection_options = {})
...
-                                        :expects => [201],
-                                        :headers => {'Content-Type' => 'application/json'},
-                                        :body => Fog::JSON.encode(request_body),
-                                        :method => 'POST',
-                                        :path => (uri.path and not uri.path.empty?) ? uri.path : 'v2.0'
-                                    })
+      response, expires = @@token_cache[{body: request_body, path: path}]
 
-      puts "response.body: #{response.body}" if user_domain=='Default2'
+      unless response && expires > Time.now
```
Ladas: should be expires < Time.now, right now it never expires :-)

but even then, I don't see this cache being cleared, so will this grow in memory?

lib/fog/openstack/models/identity_v3/projects.rb
```diff
@@ -8,8 +8,20 @@ class V3
         class Projects < Fog::OpenStack::Collection
           model Fog::Identity::OpenStack::V3::Project
 
+          @@cache = {}
+          Fog::Identity::OpenStack::V3::Project.use_cache(@@cache)
+
           def all(options = {})
-            load_response(service.list_projects(options), 'projects')
+            cached_project, expires = @@cache[{token: service.auth_token, options: options}]
```
Ladas: hm this doesn't work, I think I am able to rescope the same token, so I am somehow getting here projects that I don't have rights to.

And later in the code getting 
Expected([201]) <=> Actual(401 Unauthorized)
       excon.error.response
         :body      => "{\"error\": {\"message\": \"Could not find project: 6923ecb4505e44a9b614912ce6c7cd5b-a3dd363c-de55-4984-9f30-eec3c5a (Disable debug mode to suppress these details.)\", \"code\": 401, \"title\": \"Unauthorized\"}}"

Ladas:
@dhague for now I have tests just for project list, so the issue might be elsewhere

I am not really convinced that there should be a caching layer on this level. Seems like it can bring more problems in a future than benefits. :-) Or at least it should be configurable somehow.

lib/fog/openstack/models/identity_v3/tokens.rb
```diff
@@ -10,6 +10,7 @@ class Tokens < Fog::OpenStack::Collection
 
 
           def authenticate(auth)
+            @@cache ||= {}
```
Ladas: not sure what this does, since there is nothing in the cache?

dhague: This lazily initializes the cache to an empty Hash

lib/fog/openstack/models/identity_v3/projects.rb
dhague: Could you send me some test code that exhibits the breaking behaviour? That will make it easier for me to reproduce & fix.

lib/fog/openstack/core.rb
dhague: Not sure I agree - the logic here is: "if there is a cached response, and if the expiry time is in the future (i.e. greater than 'now'), then don't bother making a request".

dhague: Running the VCR tests, there are 5 hits on the token cache, and 25 misses - so there is some utility there.
The cache size goes as high as 13 entries, but does get cleared down and finished the test suite with 2 entries.

Ladas: @dhague ah, you are right, I missed the unless, so it's actually expires <= Time.now

lib/fog/openstack/models/identity_v3/projects.rb
Ladas: @dhague ah, so the reason is I was calling just .projects method, in this case, it is caching empty list for me

It works when I call .projects.all.

So seems like there is some lazy evaluation in place? That is calling .all method later?

lib/fog/openstack/models/identity_v3/tokens.rb
Ladas: @dhague right, the question was about, that I don't see this cache used anywhere?

lib/fog/openstack/core.rb
Ladas: @dhague hm, the problem here is that our CI is doing a lot of queries, then it's not deterministic which of those will be cached, when we record VCRs, it's almost imposible.

PR #3743, opened 2015-11-02
Fix Digital Ocean V2 ssh key creation

davebenvenuti:
The Digital Ocean V2 API returns a 201 on success when creating an ssh key, not a 202: https://developers.digitalocean.com/documentation/v2/#create-a-new-key

I've verified this works locally.  Currently, fog throws an excon exception upon successful creation of an ssh key


nirvdrum:
Looks good. Thanks!

geemus:
Thanks!

Issue #3744, opened 2015-11-02
Problem connecting to OpenStack Swift Object Storage using Auth V2.


mikaelhm:
When I try to connect to a private hosted OpenStack Swift Object Storage, using auth v2. 
I get the following error:
```
Uncaught exception: Multiple regions available choose one of these 'default'
	/Users/mikaelhm/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/fog-1.35.0/lib/fog/openstack/core.rb:266:in `authenticate_v2'
``` 

Investigating this a bit, I notice that the [`service` object](https://github.com/fog/fog/blob/master/lib/fog/openstack/core.rb#L216) has multiple `endpoints` within the same region;
```json
{
  "endpoints": [
    {
      "versionId": 1,
      "publicURL": "http://swift_private.example.com/v1/LDAP_tenant",
      "region": "default",
      "tenantId": "LDAP_tenant"
    },
    {
      "versionId": 1,
      "publicURL": "http://swift_private.example.com/v1/LDAP_username",
      "region": "default",
      "tenantId": "LDAP_username"
    },
    {
      "versionId": 1,
      "publicURL": "http://swift_private.example.com/v1/LDAP_other_tenant",
      "region": "default",
      "tenantId": "LDAP_other_tenant"
    }
  ],
  "type": "object-store",
  "name": "Swift Object Storage (http://swift_private.example.com)"
}
```

Based on the example provided in the swiftstack docs (in the bottom her): https://swiftstack.com/docs/cookbooks/swift_usage/auth.html
It looks like Fog should support multiple endpoints for a single region.

**Any thoughts on this?**

*Bonus info: This is how I connect:*

```ruby
credentials = YAML.load_file('credentials.yml')

connection = Fog::Storage.new({
  :provider            => 'OpenStack',   # OpenStack Fog provider
  :openstack_username  => credentials['username'],
  :openstack_api_key   => credentials['api_key'],
  :openstack_auth_url  => credentials['auth_url_v2'],
  :openstack_region    => 'default',
  :openstack_tenant    => credentials['tenant']
})
```

geemus:
closed by #3745

PR #3745, opened 2015-11-02
Fixed issue with OpenStack Auth v2

mikaelhm:
As mentioned in #3744 the "Multiple regions available choose one of these" exception could be raised even when there is only one region.

This happened as the check was performed directly on the array of endpoints, which might include more than one.


geemus:
Thanks!

PR #3746, opened 2015-11-02
[Rackspace] Fix set node type for Rackspace load balancer

portovep:
Fixed bug building the node type request. I also added a test for setting the node type.

I run into this issue trying to set a failover node:

https://community.rackspace.com/products/f/25/t/4377


plribeiro3000:
Sounds good. Thanks!

geemus:
Thanks!

PR #3749, opened 2015-11-03
[vsphere] support for snapshot list and revert

ireznice:


lib/fog/vsphere/models/compute/snapshot.rb
```diff
@@ -0,0 +1,62 @@
...
+        attribute :mo_ref
+        attribute :tree_node
+        attribute :snapshot_name_chain
+        attribute :ref_chain
+
+        attr_reader :service
+
+        def initialize(attributes = {})
+          super defaults.merge(attributes)
+          requires :mo_ref
```
plribeiro3000: Can't you initialize an empty object?
I suspect this `requires` could go on the actual api calls, what do you think?

lib/fog/vsphere/models/compute/snapshot.rb
```diff
@@ -0,0 +1,62 @@
...
+          @service = attributes[:service]
```
plribeiro3000: This line is not necessary. You should be able to access the `service` method already.

lib/fog/vsphere/models/compute/snapshot.rb
```diff
@@ -0,0 +1,62 @@
...
+            quiesced: false,
+            description: '',
+            create_time: nil,
+            power_state: 'none',
+            mo_ref: nil,
+            tree_node: nil,
+            snapshot_name_chain: nil,
+            ref_chain: nil
+          }
+        end
```
plribeiro3000: You should use `fog` DSL for defaults.

Like:

```ruby
attribute :quiesced, :default => false,
```

lib/fog/vsphere/models/compute/snapshots.rb
```diff
@@ -0,0 +1,31 @@
...
+          end
+        end
+
+        def get(snapshost_ref)
+          all.each do |s|
+            nested = s.get(snapshost_ref)
+            return nested if nested
+          end
+          nil
+        end
```
plribeiro3000: Isn't there an api to get only one snapshot?
You need to load all of them in memory and then sort it out?

lib/fog/vsphere/requests/compute/list_child_snapshots.rb
```diff
@@ -0,0 +1,68 @@
+module Fog
+  module Compute
+    class Vsphere
+      class Real
+        def list_child_snapshots(snapshot, opts = {})
+          s = Hash === snapshot ?
+            Snapshot.new(snapshot.update(service: self)) : snapshot
+          s.tree_node.childSnapshotList.map do |snap|
```
plribeiro3000: What the thing with this method `childSnapshotList`. Can't you use snake_case?

lib/fog/vsphere/requests/compute/list_child_snapshots.rb
```diff
@@ -0,0 +1,68 @@
...
+              item,
+              opts[:recursive] ? list_child_snapshots(item, opts) : nil
+            ]
+          end.flatten.compact
+        end
+
+        protected
+
+          def child_snapshot_info(snap_tree, parent_snap)
+            {
```
plribeiro3000: Why do you need this? Maybe you could return an actual `Snapshot` Object and let this attribute mapping to the model?

plribeiro3000:
@ireznice Thank you for submitting this. Code looks good beside some little improvements and standards you might need to fix.

lib/fog/vsphere/models/compute/snapshot.rb
ireznice: This requires is not  necessary but some methods (e.g. get, child_snapshots) would not work correctly - they depend on it.

Shall I remove it?

plribeiro3000: The i would say that this `requires` should be inside `child_snapshots` and not in the constructor.

lib/fog/vsphere/models/compute/snapshots.rb
ireznice: Unfortunately there is not. It is necessary to go recursively by calling [childSnapshotList](https://pubs.vmware.com/vsphere-60/index.jsp#com.vmware.wssdk.apiref.doc/vim.vm.SnapshotTree.html?resultof=%2522%2576%2569%2572%2574%2575%2561%256c%254d%2561%2563%2568%2569%256e%2565%2553%256e%2561%2570%2573%2568%256f%2574%2554%2572%2565%2565%2522%2520%2522%2576%2569%2572%2574%2575%2561%256c%256d%2561%2563%2568%2569%256e%2565%2573%256e%2561%2570%2573%2568%256f%2574%2574%2572%2565%2522%2520) method.

lib/fog/vsphere/requests/compute/list_child_snapshots.rb
ireznice: Sorry, it is not possible because it is a method of Vsphere API provided by rbvmomi library.

lib/fog/vsphere/requests/compute/list_child_snapshots.rb
ireznice: It is implemented the same way as other requests in vsphere (see [list_networks.rb](https://github.com/fog/fog/blob/master/lib/fog/vsphere/requests/compute/list_networks.rb) or [list_datastores.rb](https://github.com/fog/fog/blob/master/lib/fog/vsphere/requests/compute/list_datastores.rb).

Furthermore it would not be easy to implement it in this way, the properties are composed from multiple inputs: snap_tree, parent_snap.

lib/fog/vsphere/models/compute/snapshots.rb
plribeiro3000: Ok then. =)

lib/fog/vsphere/requests/compute/list_child_snapshots.rb
plribeiro3000: Damm. Is not in your hands then. Too bad the guys on `rbvmomi` made such a confusion.
The mixed up both snake_case and camelCase. =s

From their own example:

```ruby
vim.serviceInstance.find_datacenter("mydatacenter")
```

=s

ireznice:
Thank you for your comments.
Issues will be fixed ASAP, notes with no reaction will be fixed automatically.

lib/fog/vsphere/requests/compute/list_child_snapshots.rb
plribeiro3000: Ok. You are right here. But then the `vsphere` provider need a good refactoring. It seems to be using some old standards from `fog`. Lets tackle this refactoring in another PR.

plribeiro3000:
@ireznice Thank you for bearing with me on this. :+1:

ireznice:
I've fixed the snapshot.rb, tests unfortunately failed but the fail is not related to vsphere

lib/fog/vsphere/models/compute/snapshot.rb
```diff
@@ -0,0 +1,42 @@
...
+        attribute :mo_ref
+        attribute :tree_node
+        attribute :snapshot_name_chain
+        attribute :ref_chain
+
+        attr_reader :service
+
+        def initialize(attributes = {})
+          super attributes
+        end
```
plribeiro3000: You don't need to override `initialize` here.

lib/fog/vsphere/models/compute/snapshot.rb
```diff
@@ -0,0 +1,42 @@
...
+        attribute :description, :default => ''
+        attribute :create_time
+        attribute :power_state, :default => 'none'
+        attribute :ref
+        attribute :mo_ref
+        attribute :tree_node
+        attribute :snapshot_name_chain
+        attribute :ref_chain
+
+        attr_reader :service
```
plribeiro3000: You don't need to create this. In matter of effect i think it will just override the `service` method that is existent already breaking this model.

lib/fog/vsphere/models/compute/snapshot.rb
```diff
@@ -0,0 +1,42 @@
...
+
+        def initialize(attributes = {})
+          super attributes
+        end
+
+        def child_snapshots(filters = {})
+          requires :mo_ref
+          service.snapshots(
+            { :server_id => server_id, :parent_snapshot => self }.update(filters)
```
plribeiro3000: Shouldn't this method `requires :server_id`? Because you are not using `mo_ref` here but `server_id`.

lib/fog/vsphere/models/compute/snapshot.rb
```diff
@@ -0,0 +1,42 @@
...
+          )
+        end
+
+        def get(snapshot_ref)
```
plribeiro3000: I believe this method signature should be `get_childs`. `get` is kind of confusing because the collection object will have a `get` method. Does that make sense to you?

lib/fog/vsphere/models/compute/snapshots.rb
```diff
@@ -0,0 +1,31 @@
...
+            requires :server_id
+            load service.list_vm_snapshots(server_id, filters)
+          end
+        end
+
+        def get(snapshost_ref)
+          all.each do |s|
+            nested = s.get(snapshost_ref)
+            return nested if nested
+          end
```
plribeiro3000: I think you could use `fog` search feature here.

```ruby
all.find { |snapshot| snapshot.id == snapshot_ref }
```

lib/fog/vsphere/requests/compute/list_child_snapshots.rb
```diff
@@ -0,0 +1,68 @@
...
+        def list_child_snapshots(snapshot, opts = {})
+          s = Hash === snapshot ?
+            Snapshot.new(snapshot.update(service: self)) : snapshot
+          s.tree_node.childSnapshotList.map do |snap|
+            item = child_snapshot_info(snap, s)
+            [
+              item,
+              opts[:recursive] ? list_child_snapshots(item, opts) : nil
+            ]
+          end.flatten.compact
```
plribeiro3000: Could you assign the result of `s.tree_node.childSnapshotList.map` to a variable and then call `.flatten.compact` on that variable at the end of your method?
Chaining methods to block like this isn't that readable nor easy to follow.

lib/fog/vsphere/models/compute/snapshot.rb
ireznice: I agree, this method does the same as collection object (Snapshots#get). Therefore I prefer `get` signature. Even so, It could return self as well.

Shall I keep this signature? ...or do you prefer `find` or anything else instead?

plribeiro3000: This method returns the childs only, right?
So perhaps `get_childs` or something on those lines, just to be more explicit about it. It ok for the collection object to have a method `get` because is kind of our current convention but it is too much of generic for the model itself.
I feel that `find` is just renaming but the issue remains. Do you have any other ideas?

plribeiro3000: Looking again at the code i feel we don't need this method. It is kind of doing the exact same job as

```ruby
 connection.snapshots.get(snapshot_ref)
```

so i would avoid such code here. What do you think?

ireznice:
I've fixed the code, please check the last reaction.
Thank you

lib/fog/vsphere/requests/compute/list_vm_snapshots.rb
```diff
@@ -0,0 +1,64 @@
...
+    class Vsphere
+      class Real
+        def list_vm_snapshots(vm_id, opts = {})
+          get_vm_ref(vm_id).snapshot.rootSnapshotList.map do |snap|
+            item = snapshot_info(snap, vm_id)
+            [
+              item,
+              opts[:recursive] ? list_child_snapshots(item, opts) : nil
+            ]
+          end.flatten.compact
```
plribeiro3000: Can you push this to a variable as well please?

Thanks.

lib/fog/vsphere/models/compute/snapshots.rb
```diff
@@ -0,0 +1,27 @@
...
+          if parent_snapshot
+            load service.list_child_snapshots(parent_snapshot, filters)
+          else
+            requires :server_id
+            load service.list_vm_snapshots(server_id, filters)
+          end
+        end
+
+        def get(snapshost_ref)
+          all.find { |snapshost| snapshost.get(snapshost_ref) }
```
plribeiro3000: misspelling? I believe it should be `snapshot_ref`.

plribeiro3000:
@ireznice What about the `get` method at the `Snapshot` object?

lib/fog/vsphere/models/compute/snapshots.rb
ireznice: yes misspelling only, sorry

lib/fog/vsphere/models/compute/snapshot.rb
lksv: Hi @plribeiro3000 I am a colleague of @ireznice.

(do you mean `Fog::Compute::Vsphere::Real === connection`? e.g. `connection.snapshots(server_id: ...).get(snapshot_ref)`)

This method recursively goes through snapshot tree and finds/gets the snapshot by its `ref` (also self may be returned). Unfortunately it has to be done recursively by "rbvmomi" call for each snapshot node in tree.

In current implementation, `Snapshots#get` recursively calls `get` on `Snapshot`. It could be refactored in `Snapshots#get` but I think current version is slightly more readable.

Do you suggest to rewrite it to `Snapshosts`? We will fully accept your opinion.

plribeiro3000: Got it. Makes sense then. Lets just rename it to something more explicit and i think we are gold.

ireznice: Do you have some idea? It must be singular because it returns only self or some child snapshot. Do you think that get_child is OK?

Thanks.

plribeiro3000: Hmmm. get_child sound nice.

Em qua, 4 de nov de 2015 12:34, ireznice <notifications@github.com>
escreveu:

lib/fog/vsphere/models/compute/snapshot.rb
```diff
@@ -0,0 +1,35 @@
...
+        attribute :mo_ref
+        attribute :tree_node
+        attribute :snapshot_name_chain
+        attribute :ref_chain
+
+        def child_snapshots(filters = {})
+          service.snapshots(
+            { :server_id => server_id, :parent_snapshot => self }.update(filters)
+          )
+        end
```
plribeiro3000: Just noticed now. `child_snapshots` is updating the object?
I would expect such signature to only list.

plribeiro3000:
@ireznice I'm sorry for keep raising questions and points on your PR.
Your code is fine since the beginning. I'm just trying to follow some conventions and to have a good api.
Since your code is adding new stuff, the job to get it in a good shape is harder so thank you to keep bearing with me on this! :+1:

lib/fog/vsphere/models/compute/snapshot.rb
ireznice: Hello,
`child_snapshots` is not updating the object, that update method only updates the hash which is the argument to `service.snapshots`.

ireznice:
@plribeiro3000 I've changed `get` to `get_child`, please, check it out.

plribeiro3000:
LGTM! :+1:

geemus:
Thanks!

PR #3754, opened 2015-11-10
[vsphere] allow setting of boot order when using api > 5.0

timogoebel:
This PR adds the functionality to set the vmware boot order.

### DIFF
--- a/lib/fog/vsphere/requests/compute/create_vm.rb
+++ b/lib/fog/vsphere/requests/compute/create_vm.rb
@@ -19,6 +19,7 @@ module Fog
           vm_cfg[:cpuHotAddEnabled] = attributes[:cpuHotAddEnabled] if attributes.key?(:cpuHotAddEnabled)
           vm_cfg[:memoryHotAddEnabled] = attributes[:memoryHotAddEnabled] if attributes.key?(:memoryHotAddEnabled)
           vm_cfg[:firmware] = attributes[:firmware] if attributes.key?(:firmware)
+          vm_cfg[:bootOptions] = boot_options(attributes) if attributes.key?(:boot_order)
           resource_pool = if attributes[:resource_pool]
                             get_raw_resource_pool(attributes[:resource_pool], attributes[:cluster], attributes[:datacenter])
                           else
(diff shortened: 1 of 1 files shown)


lib/fog/vsphere/requests/compute/create_vm.rb
```diff
@@ -54,6 +57,54 @@ def device_change attributes
...
+              boot_order << disks.map { |disk| create_boot_device(:disk, disk, disks.index(disk)) }
+            end
+
+            if boot_device == 'cdrom'
+              boot_order << create_boot_device(:cdrom)
+            end
+
+            if boot_device == 'floppy'
+              boot_order << create_boot_device(:floppy)
+            end
```
plribeiro3000: Could you change this to a `case` statement?

plribeiro3000:
AFAICT you have a `case` statement that then calls another method with the same `case` statement.
Why can't you have it all in the same place (just 1 `case` statement)?

timogoebel:
@plribeiro3000 : Thanks for reviewing this! I refactored the code to match your suggestions. What do you think?

plribeiro3000:
Thats better. Now i can see the complexity of this code easier. Before it was not so clear all the branchs and conditionals in your code.
Perhaps you could add some tests around this code? And at least document whats happening in the `network` and `disk` case. It is clear to me now but i'm pretty sure it will slip from my mind soon.

marcohelmerich:
@plribeiro3000 
Actually, we copied the "Like why add +1 to the disk if greater then 6." part from create_disk(). I do not see any sense in it. SCSI-ID 7 is not allowed for disks as it belongs to the controller, but the code does not prevent a key with value 7. And the key-parameter das not even set the scsi id, either...

According to VmWare keys need to be immutable at creation time, need to be set by the client and are temporary. After creation the key is reset and controlled by the server. First Nic starts at 4000, First Disk starts at 2000. 
To set the boot order you need to specify the server side keys e.g. 4000 , 4001...

We can write tests and documentation. I would remove the +1 stuff completely from the code.

plribeiro3000:
@marcohelmerich Well, i know 0 for `vsphere` but if you say so i would go for it.
Your explanation seems reasonable and i would accept a code that only skip the number 7. I think this makes way more sense then the original one.

I would be happy if you fix it and add a documentation explaining this with a link to the documentation that explain this if there is one.

Thanks!

timogoebel:
@plribeiro3000: @marcohelmerich and I had a closer look at this. It seems like fog currently only supports a single scsi controller, so we based our PR on that assumption.
We added docs of our findings to the code.

Can you push me in the right direction as to how to write a test for this? If I am not mistaken, there are no tests for host creation, yet...

plribeiro3000:
@timogoebel Thank you for taking the time to add some docs to it.
The tests will be a little harder since all your logic is on top of the request so you will need to test the request to ensure this behaviour.
I'm fine without tests if you make sure this code works.
In the future i would like to start using the latest `fog` core features so this would be easier to maintain and test. But thats not something to be done in this PR. =)

timogoebel:
@plribeiro3000 : Just a quick point of order: #3762 removes vsphere functionality. Should we add our code to fog-vsphere?

plribeiro3000:
@timogoebel Please do. =)

PR #3762, opened 2015-11-13
Remove vSphere provider

jrgarcia:
Removes the vSphere provider from the fog/fog repository. This has been
moved into it's own repository at https://github.com/fog/fog-vsphere.


jrgarcia:
I'd love to have some feedback on this @geemus and @plribeiro3000. I'm pretty sure I got everything, but I'm not entirely certain. Another pair of eyes never hurts!

plribeiro3000:
@jrgarcia You need to move the file `lib/fog/bin/vsphere` to the extracted gem and remove it from here as well.

jrgarcia:
@plribeiro3000 Anything else?

plribeiro3000:
Not that i recall from the top of my head.
But i think it would be nice to have @geemus ideas here since he is doing some work towards the `net/ssh` issue and this might affect him.

plribeiro3000:
@jrgarcia Something else is missing. The build is still broken.

jrgarcia:
Yeah. I haven't had a chance to get back to it. `available?` wasn't defined in the bin file, but one of the specs tests for that. I'll try and take a look today, but life happened. :smile:

plribeiro3000:
There is 2 things i can think of. The requires in `fog` are wrong or the gem is not correctly loading `Fog::Service`.

I guess is the second because as far as i remember the first one don't need updates because the path reamains the same.

jrgarcia:
Thanks, man! That gives me somewhere to look. It sounds like the latter to me as well.

plribeiro3000:
You are welcome. Thank for all your work!

timogoebel:
@plribeiro3000 : What is the plan with this? Are you planning on moving to fog-vsphere in the next fog release? When is it expected? Or does it make sense to add my PRs still to this repo?

plribeiro3000:
@timogoebel I didn't realize that @jrgarcia fixed the build. I'm going to merge this right now so that next release will include this change.

PR #3766, opened 2015-11-16
Enforce no slash in container name

mdarby:
Ironically the docs mention it, but don't actually enforce no-slashes in a container name.


plribeiro3000:
:+1:

PR #3769, opened 2015-11-19
[opennebula] add suspend&resume support and change tests to fog way

eucher:


lib/fog/opennebula/requests/compute/image_pool.rb
```diff
@@ -18,10 +18,15 @@ def image_pool(filter = { })
               end
             end
           end
-          images.delete nil
           images
         end #def image_pool
       end #class Real
+
+      class Mock
```
geemus: For mocks that don't do anything yet it is probably better to just leave these undefined (it should automatically return a nice error in this case).

lib/fog/opennebula/requests/compute/template_pool.rb
```diff
@@ -150,6 +126,9 @@ def id
           def name
             "fogtest"
           end
+					def model
```
geemus: Could you fix the indentation level for this?

geemus:
Looking good overall, just a couple small comments above. Thanks!

PR #3774, opened 2015-11-23
Fix DigitalOceanV2::SshKey misplaced #save, #destroy and #update methods

### LINKED ISSUE #3773: Can not create SSH keys in Digital Ocean
Whenever I try to create a new SSH key on Digital Ocean (v2), I get the following error:

undefined method `save' for #<Fog::Compute::DigitalOceanV2::SshKey:0x007fab74942088>

The 'save', 'destroy' and 'update' methods are not correctly defined within 'Fog::Compute::DigitalOceanV2:: SshKey' but directly on 'Fog::Compute::DigitalOceanV2'.

mafraba:
Fixes #3773


geemus:
Good catch, thanks!

Issue #3775, opened 2015-11-23
Breaking change in mime-types-data prevents installation of fog 1.5.0


joshi4:
Configuration detail: 

* `ruby -v = ruby 1.9.3p547 (2014-05-14 revision 45962) [x86_64-linux]`
* chef-client -v = 11.16.4`
* running on ubuntu trusty 

I was installing fog through `chef` using the `chef_gem` resource and till yesterday everything was fine. 
However, recently Nov 21 or Nov 20th ( 2015) they introduced `mime-types-data` which requires ruby 2.0 and since fog depends on `mime-types >=0` we end up pulling the breaking dependency. 

I tried downloading an earlier version of mime-types that does not have the mime-types-data dependency but that doesn't work as when installing the gem it still tries to pull in `mime-types-data` 

~~~ruby 
chef_gem "fog" do
      action :install
      version '1.5.0'
 end
~~~ 

I saw a past issue where the dependency on mime-types was changed to `~> 1.16` but that was only for ruby 1.8.7 maybe ? Is a similar solution possible for ruby `1.9.x` ? 
mime-types link: https://rubygems.org/gems/mime-types

GAV1N:
My PR above specifies the latest non-breaking version of `mime-types` in the `gemspec`. I'm not sure if the maintainers will like that approach, but I figured it's worth a shot because I too have been bitten by this issue and I'd like to see it resolved quickly.

joshi4:
@GAV1N : nice I was able to modify my cookbook to download a compatible version of mime-types explicitly before installing fog and that seems to have solved the issue for me: 

~~~ruby
chef_gem "mime-types" do 
  action :install 
  version '1.25.0'  
end 

Gem.clear_paths
require 'mime-types' 
~~~ 

definitely hacky and your solution is more cleaner and better for the longterm.

plribeiro3000:
@joshi4 I suggest you to use @GAV1N solution which is the best one for this kind of issue.
You can specify it on your `Gemfile` directly since `fog` does not specify it directly and it will work out of the box. In the meantime i believe we might get @GAV1N solution into `fog` soon.

Thank you both. Since there is an open PR already i will close this so we can continue to discuss it there if needed.

PR #3776, opened 2015-11-23
Bring back compatibility with Ruby < 2.0

GAV1N:
The `mime-types` dependency isn't set to a specific version. As a result, the latest version is used, which requires `mime-types-data` in turn breaks compatibility with ruby <2.0 since ruby >=2.0 is one of its dependencies.

This PR simply specifies the latest non-breaking version of `mime-types` in the `gemspec`.

Fixes: https://github.com/fog/fog/issues/3775.

### DIFF
--- a/fog.gemspec
+++ b/fog.gemspec
@@ -75,7 +75,7 @@ Gem::Specification.new do |s|
 
   s.add_development_dependency("docker-api", ">= 1.13.6")
   s.add_development_dependency("fission")
-  s.add_development_dependency("mime-types", "~>2.99")
+  s.add_development_dependency("mime-types", "< 3.0")
   s.add_development_dependency("minitest")
   s.add_development_dependency("minitest-stub-const")
   s.add_development_dependency("opennebula")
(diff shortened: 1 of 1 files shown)


plribeiro3000:
@GAV1N Thank you! Could you specify that version as the top boundary?
Like `< 3.0` or something like that? Because that lock you made might break someone else system that are using an old version and can't upgrade right now.

geemus:
I think this should probably also just be on the mime-types requirement in fog-core instead of here (and or I should do the work to make it an optional require, similar to what we just did with net/ssh.

GAV1N:
I just implemented @plribeiro3000's suggestion before reading @geemus' comment. I don't have anything to add re: the optional require, since I'm not too familiar with the code base and I don't understand the implications.

geemus:
@GAV1N no worries. Thanks for your help, unfortunately it is probably something I'll need to do myself (since I have access all over). I'll try to land the related stuff soon.

GAV1N:
Got it, thanks @geemus.

geemus:
An initial attempt here: https://github.com/fog/fog-core/pull/170

PR #3779, opened 2015-11-26
[openstack] Fix for bad use of instance variables in a class method

dhague:
Thanks to @dLobatog for highlighting this at https://github.com/fog/fog/commit/4d5287020d961e794ac279c640c69a66238c33f7#commitcomment-14475988

### DIFF
--- a/lib/fog/openstack/core.rb
+++ b/lib/fog/openstack/core.rb
@@ -154,6 +154,13 @@ module Fog
         @port   = @openstack_management_uri.port
         @scheme = @openstack_management_uri.scheme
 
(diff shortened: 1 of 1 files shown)


dLobatog:
:+1: , the only problem was that @identity_connection needed to be defined within an instance method, now it is defined properly.

geemus:
Thanks!

dLobatog:
@geemus is this expected for 1.37 or there is going to be a 1.36.1 release in between with changes like this?

geemus:
Um, probably 1.37, though schedule-wise that theoretically would be Thursday in terms of timing.

PR #3783, opened 2015-12-01
Update API path for deleting SSH keys

borski:
The current path is incorrect (missing a '/'), resulting in a 404 whenever you try to destroy an SSH key on DigitalOcean V2.


geemus:
Good catch.

/cc @bryanl @petems (not quite a fix for the other ssh stuff, but related/helps).

PR #3787, opened 2015-12-04
Revert "If the management URL has no path, use the one in the auth URI."

seanhandley:
This reverts commit f17419049162a582268e0bace3edd4fdc17527d4.

Seems to cause some issues elsewhere in the code not caught by the current test suite but emerging from staging tests.


geemus:
Thanks!

PR #3788, opened 2015-12-04
[openstack] Workaround for installs using both Keystone v2 and v3.

seanhandley:
Introduces a new optional parameter for OpenStack.

```ruby
  openstack_identity_prefix: "v2.0"
```

This gets inserted wholesale into calls to the identity service (not including the original token request).

On systems where Keystone v2 is left in for compatibility in addition to Keystone v3 being present, it is recommended to remove the API version from the identity endpoints in Keystone's database. This means the client is then responsible for selecting which version of the API to use.

Since Keystone returns `host:5000/`, it's up to the client to add on `v2.0` or `v3`.

If left blank, the identity URL returned by Keystone's service catalogue is used unmodified. If you need to select between v2 and v3 then set it as above.


geemus:
Thanks!

PR #3793, opened 2015-12-11
vCloud Director: Add lots more mocks

pedantic-git:
This pull request adds a lot more mocks to the vcloud_director provider.

Specifically, it adds mocks for:

* delete_vapp
* get_catalog
* get_catalog_item
* get_metadata
* get_vapp
* get_vm
* get_vm_disks
* get_vm_network
* get_vms
* post_instantiate_vapp_template
* post_power_on_vapp
* post_update_vapp_metadata
* put_network_connection_system_section_vapp

And fixes a bug in the get_vm_network mock (it was not pulling out the VM ID correctly from the href - I changed the code to match the 'real' parser).

### DIFF
--- a/lib/fog/vcloud_director/compute.rb
+++ b/lib/fog/vcloud_director/compute.rb
@@ -689,6 +689,9 @@ module Fog
                 :name => org_name,
                 :uuid => uuid
               },
+              
+              :tags => {},
+              
               :tasks => {},
 
               :vapps => {
--- a/lib/fog/vcloud_director/compute.rb
+++ b/lib/fog/vcloud_director/compute.rb
@@ -506,12 +506,16 @@ module Fog
       class Mock
         attr_reader :end_point, :api_version
 
+        # This is used in some mocks so it's a method rather than a variable
+        def default_network_uuid
+          @default_network_uuid ||= uuid
+        end
+
         def data
           @@data ||= Hash.new do |hash, key|
 
             vdc1_uuid = uuid
             vdc2_uuid = uuid
-            default_network_uuid = uuid
             uplink_network_uuid  = uuid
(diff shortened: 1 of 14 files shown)


plribeiro3000:
LGTM. Need to wait for the build result.

PS: @pedantic-git Would you be interested in extract the `vcloud` provider of fog?

pedantic-git:
@plribeiro3000 Do you mean removing it from the `fog` repo and making it a repo of its own? I'm happy to do this if you like. Note: this is the `vcloud_director` provider which is different from the `vcloud` provider.

It looks like the CI has failed on all the "edge" Gemfiles. Do you know what might have caused this?

plribeiro3000:
@pedantic-git Yeap. extract it for its own gem. In the past i would do it myself but i'm not that available right now.

So i would me more then happy if you could do it. I can certainly help to review.

plribeiro3000:
Not sure whats the issue. Will take a look later.

pedantic-git:
@plribeiro3000 Awesome - yes, I'm happy to extract it into its own gem, especially if I can do it after my 4 PRs are merged! :wink:

Do you have any guidance for how to do this?

plribeiro3000:
Yes. The first step would be to only move the code and make it work the way it is. 
It should be pretty much straightforward.

After that is more like improving the code to work in a better shape because the current structure is not optimized for a standalone gem.

But we can go step by step. Once you create a new repo with the code and it work we can see about the next one.

Regarding your PR's let get them merged. =)

pedantic-git:
OK! I look forward to working with you to extract this into a separate gem.

plribeiro3000:
Me too. You can start it on your own user and we can move it later.

geemus:
Thanks!

:+1: on extracting

PR #3794, opened 2015-12-11
vCloud Director: Fix a bug in put_guest_customization_section_vapp

pedantic-git:
The vCloud Director API angrily rejects requests of this type where `:AdminPasswordAuto` is true but `:AdminPassword` is also supplied.

This PR removes the `:AdminPassword` key in those situations.


plribeiro3000:
Looking good. Will wait for the build result.

geemus:
Thanks!

PR #3795, opened 2015-12-11
vCloud Director: Add a #storage_only to disks

pedantic-git:
The standard behaviour of `vm.disks` returns all disk hardware including SCSI/IDE controllers.

This tiny PR adds a `vm.disks.storage_only` that returns only those disks that have a resource type of 17 (indicating that they're actually disks and not controllers).

### DIFF
--- a/lib/fog/vcloud_director/models/compute/disks.rb
+++ b/lib/fog/vcloud_director/models/compute/disks.rb
@@ -33,7 +33,8 @@ module Fog
           items
         end
         
-        # Filters out all the controllers (returns only resource_type == 17)
+        # Returns only disk drives (OVF resource type 17) and not controllers,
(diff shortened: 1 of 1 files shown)


lib/fog/vcloud_director/models/compute/disks.rb
```diff
@@ -32,6 +32,11 @@ def item_list
           end
           items
         end
+        
+        # Filters out all the controllers (returns only resource_type == 17)
```
plribeiro3000: Can you be more explicit here and explain what a disk of `resource_type` equals 17 is?

pedantic-git: Thanks for the suggestion. I've added a comment saying it's the OVF resource type and a link to a VMware blog on the subject.

geemus:
Thanks!

PR #3796, opened 2015-12-11
vCloud Director: Add support for reconfigureVm and a VM's description

pedantic-git:
Sorry for raising a PR with two purposes here, but they're quite entwined in my fork.

1. Adds support for reading the `Description` property of VMs.
2. Adds some minimal support for the all-encompassing `reconfigureVm` method of the vCloud Director API. Specifically, you can use the `vm.reconfigure` method to set the following properties:

* `name`
* `cpu`
* `memory`
* `description`

[I know there is already support for PUT `cpu` and PUT `memory` but this method allows people to set them both (and other properties) simultaneously so it's a bit faster. The other two properties previously had no way of being edited in Fog.]

### DIFF
--- a/lib/fog/vcloud_director/compute.rb
+++ b/lib/fog/vcloud_director/compute.rb
@@ -225,6 +225,7 @@ module Fog
       request :post_power_off_vapp
       request :post_power_on_vapp
       request :post_reboot_vapp
+      request :post_reconfigure_vm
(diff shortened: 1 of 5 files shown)


plribeiro3000:
LGTM!

geemus:
Thanks!

PR #3797, opened 2015-12-11
[vcloud_director] Add vdc.networks

pedantic-git:
As well as `org.networks` which gets all the networks in an organization, it is useful to have `vdc.networks` that only gets the networks relevant to a specific virtual DC.


geemus:
Thanks!

Issue #3798, opened 2015-12-15
[digitalocean] server destroy throws NoMethodError


rroa:
Even though the request is defined, the destroy method doesn't seem to be implemented for the server model.

```
undefined method `destroy' for #<Fog::Compute::DigitalOceanV2::Server:0x007fce7c8b2728> 
(NoMethodError)
```

h0lyalg0rithm:
Hi @rroa, to delete the server from digitalocean you have call the delete method.

h0lyalg0rithm:
@geemus I think you can close this one :)

h0lyalg0rithm:
@rroa I realized that the example in the repo was wrong.Thanks for letting us know.

PR #3803, opened 2015-12-16
Update getting_started

h0lyalg0rithm:
Changed the example to use the new method. #3798

### DIFF
--- a/lib/fog/digitalocean/examples/getting_started.md
+++ b/lib/fog/digitalocean/examples/getting_started.md
@@ -120,5 +120,5 @@ server.power_cycle
 Destroying the server:
 
 ```ruby
-server.destroy
+server.delete
 ```


geemus:
I think, actually, that we should change the server model to have a `destroy` method. That is what is expected for all models in fog and probably just got accidently overlooked here. Would you be up for adding the method here to do that? We should maybe also add a deprecation warning around using delete and have it call destroy. What do you think?

h0lyalg0rithm:
Same here, its better to be consistent with the convention.
Deprecation warning would also be good.Maybe I should revert the change on the readme.

geemus:
Sounds like a plan, just let me know if you have questions or there is
anything I can help with. Thanks!

On Mon, Dec 21, 2015 at 11:22 PM, Suraj Shirvankar <notifications@github.com

h0lyalg0rithm:
@geemus Was a bit busy the past couple of days.Will try to send a PR soon hopefully before christmas.

geemus:
No worries, thanks for the update.

On Wed, Dec 23, 2015 at 10:26 AM, Suraj Shirvankar <notifications@github.com

PR #3804, opened 2015-12-16
Added method locked?

h0lyalg0rithm:
checks if the digitalocean droplet is currently locked with another operation


geemus:
Thanks!

PR #3805, opened 2015-12-18
Add openstack_temp_url_key to the list of recognized options

snoopie:
Hello guys, I noticed a warning when using ```:openstack_temp_url_key``` option for OpenStack storage which doesn't make much sense:
```
2015-12-18_08:11:07.54954 [fog][WARNING] Unrecognized arguments: openstack_temp_url_key
```

when in reality code still uses it and it is perfectly valid. See https://github.com/fog/fog/blob/master/lib/fog/openstack/requests/storage/get_object_https_url.rb#L39, https://github.com/fog/fog/blob/master/lib/fog/openstack/requests/storage/get_object_https_url.rb#L56.

I just whitelisted it as a recognized one. Please have a look and let me know if you have any questions.

Thanks!


geemus:
Thanks!

PR #3806, opened 2015-12-29
[OpenStack] service retrieval also pass options on unscoped fallback

MaPoCaCe:
Same as in [line 307](https://github.com/MaPoCaCe/fog/blob/9589cad1c6332427415fb91ca49e8ae8af732c5b/lib/fog/openstack/core.rb#L307) the options should be passed to the service retrieval.

FYI: I stumbled upon this, when I needed the option `openstack_endpoint_path_matches` passed to select the correct identity version.


geemus:
Thanks!

PR #3807, opened 2015-12-29
Deprecate delete_server for destroy_server

h0lyalg0rithm:
Ref #3803 
Use destroy_server to maintain consistent destroy api.

### DIFF
--- a/lib/fog/digitalocean/compute_v2.rb
+++ b/lib/fog/digitalocean/compute_v2.rb
@@ -21,7 +21,7 @@ module Fog
       request :change_kernel
       request :create_server
       request :create_ssh_key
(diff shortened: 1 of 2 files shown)


h0lyalg0rithm:
@geemus I build seems to be broken 
```
Gem::InstallError: mime-types-data requires Ruby version >= 2.0.

An error occurred while installing mime-types-data (3.2015.1120), and Bundler
```

Looks this dependency requires ruby 2.0 and above so are we dropping support for ~1.9

geemus:
Thanks! I've been trying to keep mime-type support by making it not required, but haven't worked out all the kinks.

PR #3809, opened 2015-12-30
[OpenStack] image v2 correct images find_by_* and find_attribute

MaPoCaCe:
- fix what most probably is a copy paste error evolving from v1
- v2 only has the `list_images` request that accepts hash arguments only


geemus:
Thanks!

PR #3810, opened 2016-01-02
[DigitalOcean] Added filtering capabilities to images request.

rroa:
* Added filtering hash to list_images request.
* Appended filter query to list images request url
* Passed existing filtering hash from all method at images collection model to the request call.

Current query does not show all available images. In addition the **all** method at the images does not pass the filters param to the request query. The behavior of the method differs from the API one since it's just the default call.


geemus:
Thanks!

PR #3816, opened 2016-01-06
[openstack] storage does not require api_key and username

MaPoCaCe:
- same as with all the other services `openstack_api_key` and `openstack_username` are not required
- enables using the `openstack_auth_token`

### DIFF
--- a/lib/fog/openstack/storage.rb
+++ b/lib/fog/openstack/storage.rb
@@ -3,8 +3,7 @@ require 'fog/openstack/core'
 module Fog
   module Storage
     class OpenStack < Fog::Service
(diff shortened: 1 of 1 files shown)


geemus:
Thanks!

MaPoCaCe:
Thanks for the quick merge. May I ask, why you accept it though the tests are failing? Is it okay in the transition to Ruby version >= 2?

geemus:
@MaPoCaCe yeah, I checked the tests but they appeared to be unrelated (and a known issue I've been trying to work on). Hopefully that will get fixed soon. We've been trying to avoid getting rid of old rubies since many still use them, but it will inevitably happen eventually.

Issue #3777, opened 2015-11-24
Digital Ocean's list_images request does not return the full list of images 


mafraba:
It seems that by default the Digital Ocean API uses pagination, so it returns a page with a few images and a link to the next.

Right now the list_images request is just returning the images that come in the first page, but I guess this is not the behavior one would expect.

geemus:
@bryanl could you take a look when you have a chance? Thanks!

bryanl:
@mafraba I'll take a look at this.

angelman:
Any updates? Thanks. :purple_heart:

geemus:
I merged https://github.com/fog/fog/issues/3777#issuecomment-190408860 earlier today, I think that should resolve this actually. Do let me know if that doesn't seem to be the case though.

PR #3822, opened 2016-01-15
[DigitalOcean] Added filtering capabilities to servers request.

rroa:
#3777 in the same way images were updated to support filtering, here's the same update for servers.

### DIFF
--- a/lib/fog/digitalocean/models/compute_v2/servers.rb
+++ b/lib/fog/digitalocean/models/compute_v2/servers.rb
@@ -15,7 +15,7 @@ module Fog
         # @raise [Fog::Compute::DigitalOceanV2::ServiceError]
         # @see https://developers.digitalocean.com/documentation/v2/#droplets
         def all(filters = {})
-          data = service.list_servers.body['droplets']
+          data = service.list_servers(filters).body['droplets']
           load(data)
         end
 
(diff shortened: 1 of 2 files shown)


geemus:
Could you update the method signature for the mock version also? If you can update the mock version to do the right thing there that would be even better, but the difference in signature is likely to cause problems, so at least fixing that would be good. Thanks!

rroa:
Sure, will do.

rroa:
@geemus I'm having trouble finding the mocks you're referring to. I don't have much experience with shindo, as I have two PR's waiting for a similar update could you point me on the right direction for this one? 

I assumed it was at **tests/digitalocean/requests/compute_v2/list_serversL:69** but wanted to make sure it was just doing something like `tests('#list_servers(#{filters})')`

geemus:
@rroa actually, I just meant in the request definitions (I think the tests should be fine). In particular in the request files, there should be two defined methods (a real version and a mock version). So the method signature I was referring to updating, in this case, would be here: https://github.com/rroa/fog/blob/bugfix/do_servers_filters/lib/fog/digitalocean/requests/compute_v2/list_servers.rb#L16

Hope that clarifies, but just let me know if you have further questions or concerns. Thanks!

rroa:
@geemus  Thanks for the help. I have updated the signature at the Mock to match the new changes.

geemus:
Great, thanks!

PR #3823, opened 2016-01-15
[DigitalOcean]  Added filtering capabilities to regions request.

rroa:
#3777 in the same way images were updated to support filtering, here's the same update for regions.


rroa:
@bryanl @geemus

geemus:
Could you update the mock stuff here as well? If you get it all to do the filtering stuff that would be great, but we should at least match up the method signatures. Thanks!

rroa:
Sure, will do.

rroa:
@geemus Updated!

geemus:
Great, thanks!

PR #3824, opened 2016-01-18
Strip port number from host header

tlawrence:
Excon by default adds the default port to the host header for each request
e.g hostname.example.com:443

Some Cloud Providers use filter rules on HAProxy or similar to control API paths to different backends. In our case the ruleset required will increase if we need to filter for port number as well as host headers where this is not specified. 

rfc2616 specifies that the port number should be appended to the host header if it is not the default port. However, 443 IS the default port for https (as specified in rfc2818) Although I accept that this could be seen as ambiguous given that the rfc2818 refers to HTTP over TLS rather than HTTP as defined in rfc2616.

Basically what I'm trying to say (in a long winded rounadbout way..) is should we make ":omit_default_port" (as supported by Excon) the default for vCloud Director  connections? 
If so I think this PR covers it whilst allowing users to override manually if required.

### DIFF
--- a/lib/fog/vcloud_director/compute.rb
+++ b/lib/fog/vcloud_director/compute.rb
@@ -30,6 +30,7 @@ module Fog
         PORT        = 443
         SCHEME      = 'https'
         API_VERSION = '5.1'
+        OMIT_DEFAULT_PORT = true
       end
 
       class ServiceError < Fog::VcloudDirector::Errors::ServiceError; end
@@ -337,6 +338,7 @@ module Fog
           @vcloud_director_password = options[:vcloud_director_password]
           @vcloud_director_username = options[:vcloud_director_username]
           @connection_options = options[:connection_options] || {}
+          @connection_options[:omit_default_port] = Fog::Compute::VcloudDirector::Defaults::OMIT_DEFAULT_PORT unless @connection_option[:omit_default_port]
           @host       = options[:vcloud_director_host]
           @path       = options[:path]        || Fog::Compute::VcloudDirector::Defaults::PATH
           @persistent = options[:persistent]  || false


geemus:
Seems fine. More broadly I'd like to make not including the port the default in excon itself, in time. I'm afraid it might break a lot of stuff though (aws signing for sure).

geemus:
Thanks!

pedantic-git:
This PR has a typo in it (`unless @connection_option[:omit_default_port]` instead of `unless @connection_options[:omit_default_port]`) that causes initialization of `Fog::Compute::VcloudDirector` to fail in the common case. I can patch this downstream but figured it's something that you'd probably like to fix urgently upstream too.

geemus:
Thanks for the heads up, fixed here: https://github.com/fog/fog/commit/c417f0b12d4c6becd0e6b28f0a2414fd614e7c41

pedantic-git:
Thanks!

robinbowes:
Did you fix the tests too to make sure this doesn't regress? </troll>

geemus:
Not yet, figured getting this in before we forgot was a priority. I'd certainly welcome help on getting tests up to snuff also.

tlawrence:
sorry guys.. my bad!

How come the tests didn't fail?

pedantic-git:
That's a bit of a surprise to me too - they [definitely ran](https://travis-ci.org/fog/fog/jobs/103157595) (search for `VcloudDirector`).

robinbowes:
Oooh, it seems github stripped ```</troll>``` from my original comment, ie. I was being not a little facetious

geemus:
It is such a particular thing, probably just not tested appropriately. Also, since we run tests in mock mode, even if that path was exercised it would take some work to catch it. Maybe more trouble than it's worth to worry about (it seems unlikely the typo would be re-introduced). Still, definitely wish we had better coverage in general.

PR #3825, opened 2016-01-18
[DigitalOcean] Support pagination on the api

### LINKED ISSUE #3777: Digital Ocean's list_images request does not return the full list of images

h0lyalg0rithm:
Fixes #3777  
To get the items from the next link
docean.servers.next_page
docean.servers.previous_page


lib/fog/digitalocean/models/paging_collection.rb
```diff
@@ -0,0 +1,43 @@
+module Fog
+  module Compute
+    class DigitalOceanV2
+      class PagingCollection < Fog::Collection
+
+        attribute :next
+        attribute :last
+
+        def next_page
+          binding.pry
```
geemus: I think maybe this is leftover from coding/debugging?

lib/fog/digitalocean/models/paging_collection.rb
```diff
@@ -0,0 +1,43 @@
...
+        end
+
+        def get_page(link)
+          if match = link.match(/page=(?<page>\d+)/)
+            match.captures.last
+          end
+        end
+
+        def get_paged_links(links)
+          binding.pry
```
geemus: This too?

geemus:
Looks good overall, but appears to be a bit of leftover debug stuff (commented above), could you fix that up and then we can hopefully get this in?

geemus:
Thanks!

h0lyalg0rithm:
Sorry missed that wow :+1:

h0lyalg0rithm:
@geemus what do you think about the dig method, since its only available in ruby 2.3.

geemus:
@h0lyalg0rithm good question, I totally overlooked that. We are still trying to support all the way back to at least 1.9, so probably should avoid it, unfortunately.

h0lyalg0rithm:
@geemus I kinda re-implemented it so it should be fine.I was only worried that the method name was same as the http://ruby-doc.org/core-2.3.0/Hash.html#method-i-dig but the signature is different, might confuse someone in the future

geemus:
ah, ok, thanks for clarifying. Maybe we should go with a different name then, or at least a comment?

PR #3827, opened 2016-01-19
[DigitalOcean] Delete V1 api and test

h0lyalg0rithm:
- Delete compute v1
- Change to digitalocean V2 api


h0lyalg0rithm:
@geemus It might a big PR but i felt deleting code would be much better than going at it very time and having the build break.

geemus:
Thanks, yeah, since it doesn't work any more anyway, might as well clean it up.

PR #3833, opened 2016-01-28
[openstack] Add get_server_password support to openstack compute API

thindiyeh:
I added the get_server_password functionality to get the password of the instances from openstack Compute API v2.1


thindiyeh:
Please accept this PR, because it is a requirement of a new winrm support on the chef-provisioning-fog openstack driver

geemus:
Thanks!

Issue #3836, opened 2016-02-05
mock in fog openstack snapshot 'list_snapshots' error


zhitongLBN:
i got a error while using: `.snapshot.all` when i'm in the Mock mode
cause in request/compute/list_snapshots.rb
it call `get_snapshot_details.body['snapshot']` but `get_snapshot_details` need a `snapshot_id` argument.

geemus:
Ah, probably just an oversite as the interface as evolved. Would you be up for trying to work on a fix? Thanks!

zhitongLBN:
Please look at my pull request and tell me what should I improve, thx :)

geemus:
@zhitongLBN looks good, thanks!

PR #3837, opened 2016-02-08
rework mock openstack snapshot to have same behavior as servers...

zhitongLBN:
should solve issue  #3836
* Making openstack mock snapshots functions behave the same way as servers
* Solve the problem that calling `get_snapshots` but not giving `snapshot_id` in `list_snapshots` function
* Now should have the mock `volume` in data before to create mock volume `snapshot`(cause it'll have to find the size of volume that it try to snapshot)


geemus:
Thanks!

PR #3838, opened 2016-02-10
[openstack] Removes URI::encode call on parameteres

petrblaho:
which causes NoMethodError in ruby's uri/rfc2396_parser.rb
when passed parameter's value is ie. Fixnum.


geemus:
@fog/fog-openstack Could you review? Thanks!

Ladas:
@geemus looks good, seems like the encoding is breaking things and it's not really needed. It's just on those few places plus few in service actions.

CI failure seems unrelated

geemus:
@Ladas thanks, yeah, I really need to get the CI in order again.

geemus:
Thanks!

PR #3843, opened 2016-02-12
OpenStack expose header information in image download

Ladas:
We need some header information like checksum, Accept-Ranges etc

Also exposing the Range header, for partial downloads

### DIFF
--- a/lib/fog/openstack/requests/image_v2/download_image.rb
+++ b/lib/fog/openstack/requests/image_v2/download_image.rb
@@ -5,13 +5,22 @@ module Fog
         class Real
           def download_image(image_id, content_range=nil, params) # TODO: implement content range handling
             request_hash = {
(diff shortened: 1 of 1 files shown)


Ladas:
@roliveri please try if this works for you. On my OpenStack kilo, it probably doesn;t support ranges. It shoudl return in header Accept-Ranges: bytes. Maybe some Glance V2 configuration is needed first?

The example call for partial get with header returned should look like
body, header = image.download_data(:return_header => true, :content_range => "bytes=0-100")

jjasghar:
Kick this over to fog/fog-openstack?

plribeiro3000:
Good point @jjasghar. =)

plribeiro3000:
Closing in favor of fog/fog-openstack#35

PR #3845, opened 2016-02-15
[DigitalOcean] - Failed requests fail to log because of Bluebox

dLobatog:
If you're just loading fog/digitalocean in your application and some
request fails, no logs will be provided. Instead an error 'uninitialized
constant Fog::Compute::Bluebox' will show up. This is because for some
reason DO calls Bluebox::NotFound when a request fails.


geemus:
Thanks, old copy/paste error I imagine.

PR #3846, opened 2016-02-16
return network in mock data in stead of creating a new one

zhitongLBN:


geemus:
Could you give a bit more context around the purpose and expectations for this change? Thanks!

zhitongLBN:
yeah, because we are using this in our unit test, so as it create a new network, our unit test won't get pass ;-(

```
network = @connection.networks.get(floating_network_id)
if !network.nil? && network.router_external
...do sth here
```
ps: ( we use mock in travis, if not it's tooo expensive...)

zhitongLBN:
and also the code before 
```
           if data = self.data[:networks][network_id]
             response.status = 200
+            response.body = { 'network' => data }
-              'network' => {
-                'id' => 'e624a36d-762b-481f-9b50-4154ceb78bbb',
-                'name' => 'network_1',
-                'subnets' => [
-                  '2e4ec6a4-0150-47f5-8523-e899ac03026e'
-                ],
-                'shared' => false,
-                'status' => 'ACTIVE',
-                'admin_state_up' => true,
-                'tenant_id' => 'f8b26a6032bc47718a7702233ac708b9',
-              }
-            }
             response
           else
```
is quite wierd: it check if we have the `network` with `network_id in` mock data, and then it found it and return a new one instead of returning what it found

zhitongLBN:
Hi @geemus ~ please valide this PR, if you are agree with my changes :)

geemus:
Hey, sorry about the delay, have been a bit swamped at work. Just let me know if you would like to discuss this more and I'll try to be more prompt.

zhitongLBN:
Yeah sure, i thought this branch has been merged so i've deleted it.

I think this change will not have too much impact with other code exist.

geemus:
The one question I had around usage stuff, I think in many cases with something like this the expectation would be that in the mocks you would create a network and then interact with it afterward. Is this just a case where you don't really create it and it always exists in the actual system?

zhitongLBN:
I doubt that you can use `get_network` properly if you don't modify this function, cause:
in `network.py` https://github.com/fog/fog-openstack/blob/master/lib/fog/openstack/network.rb#L142-L150
you have a random `network_id` and in https://github.com/fog/fog-openstack/blob/master/lib/fog/openstack/requests/network/get_network.rb#L17 
you have to enter the right uuid to have get a network.
so every time you have to do a `networks.all` before using `get`.
At last you get another network, and you can't never edit the one already in `self.data`....
+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
what you can do(or I can do for you) is put the network which is in https://github.com/fog/fog-openstack/blob/master/lib/fog/openstack/requests/network/get_network.rb#L20-L30 into self.data, so that your "hard code" part will continue to work as before :)

geemus:
Got it, thanks. Sounds good I think, unfortunately it looks like there are merge conflicts here now. Could you rebase and I'll try to get this in? Thanks!

zhitongLBN:
yeah sure, but now the openstack is apart from gem fog, should i address this PR to the new repository?? (but it seems that the new openstack repo is no up to date, am i wrong?) 
what should I do?

geemus:
@zhitongLBN yeah, it should go in the new repo. It should hopefully be up to date (and if it isn't, please let us know and we'll try to get it all updated ASAP).

plribeiro3000:
@zhitongLBN Could you reopen this against `fog-openstack` please?

Thanks!

PR #3849, opened 2016-02-16
Standardize the `:state` alias to `:status`

jjasghar:
It seems that every other plugin expect `:state` for instance, https://github.com/fog/fog/blob/master/lib/fog/openstack/models/compute/server.rb#L34. DO needs to have the alias of `:state` to `:status` to help some down stream usage. 

Consistancy is the key to happiness :)


geemus:
Good call, thanks!

PR #3851, opened 2016-02-19
Added label to Guest Properties

tlawrence:
The current method for setting GuestProperties on vApps & VMs does not set the <ovf:Label> tag. 
this means that via vCloud Director UI Guest Property Key names are not shown, only values.

I've added the tag and tested.


geemus:
Thanks!

PR #3853, opened 2016-02-23
Keeping the dependency gems in order

jjasghar:
`fog-aliyun` was at the bottom of the list when it should be at the top,
per the comment.


plribeiro3000:
Thanks!

geemus:
Thanks!

PR #3854, opened 2016-02-23
[HPCloud] Removing HP cloud

jjasghar:
Per http://www.hpcloud.com/ the HP public cloud has been shutdown. There
still is the private cloud, but with the amount of changes that are
required and updated per instance, it would be easier to create plugins
for each.

Goodbye HP Cloud, we'll miss you.


plribeiro3000:
Does this mean any of the apis supported here work anymore?

plribeiro3000:
If so we might need to update the website as well.

jjasghar:
@plribeiro3000 yep, that is correct. The HP Public Cloud is gone gone. We (Chef) removed our knife plugin too. https://github.com/chef/knife-hp

plribeiro3000:
Nice. Thanks you for the PR then.
Could you open a PR against https://github.com/fog/fog.github.com and remove any references about it as well please?

Would be also nice to have something in the website and the README explaining it. Could you do this?

jjasghar:
@plribeiro3000 consider it done. I'll have the PR and reference it here when it's completed.

plribeiro3000:
Awesome! Thank you very much!

geemus:
@jjasghar thanks! Going to go ahead and bring this in.

jjasghar:
@geemus & @plribeiro3000 fyi: https://github.com/fog/fog.github.com/pull/47

PR #3859, opened 2016-02-29
[VCLOUD_DIRECTOR] add builder gem dependency and require

tlawrence:
instantiating vApp templates uses Builder for the XML. This is not required (and therefore throws a namespace error) and the Gem is not included in the gemfile.
Added both and seems to work now. Surprised this slipped through previously though. Guess someone had builder installed already!

### DIFF
--- a/fog.gemspec
+++ b/fog.gemspec
@@ -46,6 +46,7 @@ Gem::Specification.new do |s|
 
   s.add_dependency('nokogiri', '~> 1.5', '>= 1.5.11')
   s.add_dependency('ipaddress', '~>0.5')
+  s.add_dependency('builder', '>=3.2.2')
 
   # Modular providers
   s.add_dependency("fog-brightbox")
--- a/lib/fog/vcloud_director/requests/compute/instantiate_vapp_template.rb
+++ b/lib/fog/vcloud_director/requests/compute/instantiate_vapp_template.rb
@@ -1,3 +1,5 @@
+require 'builder'
+
 module Fog
   module Compute
     class VcloudDirector


geemus:
It is surprising that went as long as it did. Could we do more like the "optional" requires for mime-type here? Basically it will try to require it and tell you to add it nicely if you try to use it and don't have it? I suspect many fog users won't actually need builder, so adding it for everybody seems heavy handed and the "optional" requires leave something to be desired, but aren't too bad. Once we split out vcloud to it's own gem we could be safer to add this in. Sound reasonable? Oh also, looks like there are merge conflicts here, could you rebase? Thanks!

tlawrence:
sure, can you point me in the direction of an example?

geemus:
Definitely, basically like this: https://github.com/fog/fog-aws/blob/13bda13933711c7d3efa00bcf1e913e053aa34e6/lib/fog/aws/storage.rb#L141

The main part being:

```ruby
            begin
              require 'mime/types'
            rescue LoadError
              Fog::Logger.warning("'mime-types' missing, please install and try again.")
              exit(1)
            end
```

Also, we'll probably still want it in the gemspec, but just in the development dependency section. Hopefully that clarifies everything, but just let me know if you still have questions or concerns. Thanks!

tlawrence:
Hmm, I think I personally prefer to just let it error rather than bloat the code in that file. It's prettty obvious what the problem is. Is that ok with you?
I'll add to the dev dependency though.

Will close this and reopen another.

geemus:
Sure, I'm fine either way.

PR #3862, opened 2016-03-01
[VCLOUD DIRECTOR] Enhanced catalogue & vApp Template functionality

tlawrence:
virtual machines within a catalog_item are actually contained within a vAppTemplate object. This PR extends the Fog model to allow the vms inside a catalog_item to be accessed:

e.g

catalog_item.vapp_template.vms

This PR is the first step to being able to modify a VM during deployment. 

As an example:
For a template uploaded to vCloud, the network maybe preset to "VM Network". Usually you will want to reconfigure this network adapter on deployment so that the network name reflects the networks available in the VDC. To do this you need to access the HREF of the underlying VMs in the catalog_item. 

this can now be achieved via:

catalog_item.vapp_template.vms[0].href

etc..

Other changes are:

- Removal of Builder as a dependency (not used by anything else)
- Implementation of a Generator for vApp Instantiate parameters
- Refactored sourced Items XML so that VM names can be customized when deploying a vApp
- Added a minitest task for vCloud Director (dont know if its right but seems to follow the convention roughly)


geemus:
Looks good overall, but I don't think we want to add builder as a dependency for everybody? What do you think?

tlawrence:
I'll have another look at it. I don't know why builder has been used actually. Most of the generators use Nokogiri which is already there. Seems silly to have 2 x XML Builders. I'll look to see how hard it is to refactor.

geemus:
Sounds good, just let me know if there is anything I can help with. (I also
don't know why builder is used)

tlawrence:
Thanks, the only thing I need help with really is setting up Minitest. I'd like to build a generator and would prefer to use TDD to do it. I'll do some Googling and see if I can figure it out first though :-)

geemus:
Sounds like a plan, thanks!

tlawrence:
This is ready for another review I think @geemus

geemus:
Looks good, unfortunately it looks like we ended up with some test failures. Could you take a look at those? Thanks!

tlawrence:
Hi, Made a few more changes but the tests are failing for something I don't think I broke now. Unless I introduced a problem switching mocking off and back on here:

https://github.com/tlawrence/fog/blob/master/spec/vcloud_director/requests/compute/instantiate_vapp_template_spec.rb


Can you offer any advice?

geemus:
Thanks! Looks like the other errors are unrelated...

PR #3864, opened 2016-03-22
[docker] Fixing bad namespace on errors

daviddavis:
Errors don't have an "Error" namespace so we're getting unintialized constant exceptions. See https://github.com/fog/fog-core/blob/d18398c44e9e8d57025874ed43a573434dbd0f95/lib/fog/core/errors.rb.


plribeiro3000:
Awesome. Thanks!

geemus:
Thanks!

PR #3868, opened 2016-03-28
Removed 'filters' from ssh_key all method

jjasghar:
- There needs to be no filter with the all method call.
- Removed an extra WS


geemus:
Thanks!

Issue #3876, opened 2016-04-14
Fog vs. Metagems recommendation


Temikus:
Do we actually plan to continue support of fog/fog as "require everything" gem? I'm asking since some projects still use 'fog' to support only a specific provider (like ManageIQ with google) and this causes long require times and a ton of unnecessary dependencies to be loaded.

If not and we want people to use the metagems directly, maybe we should make a notice of that during install/in README, like:
NOTICE: You shouldn't require 'fog' gem directly except for:
- Supporting Ruby1.9
- Using a provider that doesn't have a metagem.

geemus:
Sounds reasonable. I think we want to continue to do the metagem thing for now for backwards compatibility at least until 2.0. Maybe once we get everything split out the meta-gem can just be retired though and we can just focus on the individual things.

connorshea:
Is it possible right now to just use the specific depencies we need, or is the metagem necessary until 2.0 is released? I was looking into how to remove extra depencies in GitLab and fog seems like a good place to start.

Temikus:
@connorshea AFAIK as long as your provider is separated to a metagem and you don't need to support Ruby1.9.3, you **should** be using the specific metagem you need.

This is precisely what I was talking about...

@geemus - how about a `spec.post_install_message`? 

With something like: 
```
------------------------------
Thank you for installing fog!

IMPORTANT NOTICE:
If there's a metagem available for your cloud provider, e.g. `fog-aws`,
you should be using it instead of requiring the full fog collection to avoid 
unnecessary dependencies.

'fog' should be required explicitly only if:
- The provider you use doesn't yet have a metagem available.
- You require Ruby 1.9.3 support.
------------------------------
```

connorshea:
@Temikus [Yeah...](https://github.com/gitlabhq/gitlabhq/blob/master/Gemfile.lock#L219)

plribeiro3000:
@Temikus You have it already then. Would you have time to send a PR with that since is your idea?

Temikus:
@connorshea Looking diagonally through the code it looks to me like you have only AWS to worry about: [1](https://github.com/gitlabhq/gitlabhq/blob/91fa250038e9182988319f088fb84741b6e2efc9/config/initializers/carrierwave.rb),[2](https://github.com/gitlabhq/gitlabhq/blob/95c7aa62defdbc7984c71faeb0b901a7fe59b223/doc/raketasks/backup_restore.md) I would double check to make sure though ;)

[...]

PR #3877, opened 2016-04-14
Adding postinstall message

Temikus:
Adding a simple postinstall message + README section to tell users not to use
fog gem if they require only specific dependencies.

Here's how it looks like:
```
λ gem install --local ./fog-1.38.0.gem                                                                                                                                                                                     (1)
------------------------------
Thank you for installing fog!

IMPORTANT NOTICE:
If there's a metagem available for your cloud provider, e.g. `fog-aws`,
you should be using it instead of requiring the full fog collection to avoid
unnecessary dependencies.

'fog' should be required explicitly only if:
- The provider you use doesn't yet have a metagem available.
- You require Ruby 1.9.3 support.
------------------------------
Successfully installed fog-1.38.0
Parsing documentation for fog-1.38.0
Installing ri documentation for fog-1.38.0
Done installing documentation for fog after 18 seconds
1 gem installed
```

Fixing #3876

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,19 @@ fog is the Ruby cloud services library, top to bottom:
 [![Code Climate](https://codeclimate.com/github/fog/fog/badges/gpa.svg)](https://codeclimate.com/github/fog/fog)
 [![Gem Version](https://badge.fury.io/rb/fog.svg)](http://badge.fury.io/rb/fog)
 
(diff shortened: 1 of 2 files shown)


plribeiro3000:
Thanks man!

geemus:
Awesome, thanks!

PR #3879, opened 2016-04-19
Update compose_common.rb

tinguelyp:
When you use recompose to clone a vm you don't have to write the following elements in xml: 
 ```
 <Description/>
  <InstantiationParams>
    <NetworkConfigSection>
      <ovf:Info/>
    </NetworkConfigSection>
  </InstantiationParams>
```
So it's better when InstantiationParams and Descripition fields are optionnal


geemus:
Thanks!

PR #3881, opened 2016-04-25
Vcloud-director create-snapshot for vm/vapp

tinguelyp:
I added two classes:
- lib/fog/vcloud_director/generators/compute/create_snapshot.rb
- lib/fog/vcloud_director/requests/compute/post_create_snapshot.rb
I also added the reference of the request in lib/fog/vcloud_director/compute.rb


geemus:
Thanks!

PR #3885, opened 2016-05-18
Update dns.rb

raravena-coupa:
Add a method to override the DNSMadeEasy connection and use the V2.0 API


raravena-coupa:
Closing... Basically I figured that this needs more changes.

PR #3886, opened 2016-05-18
Support for different API versions on DNSMadeEasy

raravena-coupa:
Creating pull request to support multiple API version in dnsme.


raravena-coupa:
Looks like this is not going to work, since I'm looking at the API docs and they changed the URLs from `domains` to `managed` and possibly others.

geemus:
Thanks for the update. Just let us know if there is anything we can help with or you want to discuss further.

PR #3891, opened 2016-06-25
Added user data as option when create digital ocean droplet

luaysuarna:


Temikus:
@geemus @plribeiro3000 I see that the fog-digitalocean was separated: https://github.com/fog/fog-digitalocean, however, it does still seem to be in `fog`. Is this WAI?

plribeiro3000:
@Temikus It never went live. If you take a look at [fog gemspec](https://github.com/fog/fog/blob/master/fog.gemspec) you will realize is not a dependency yet.

Also, the gem has only one commit so far. So i guess is wip and if stuff keep being pushed here we should get rid of that repo since the sync process is a real PITA.

cc/ @jjasghar Wanna chime in?

Temikus:
Since digitalocean provider got separated into a Gem, I'm porting this issue there: https://github.com/fog/fog-digitalocean/issues/13

@luaysuarna, please, consider rebasing your changes to the new gem.

PR #3893, opened 2016-07-07
Digital Ocean request filtering fixes & test fixes 

bsedat:
Fixes the filtering for the `list_ssh_keys`, and `list_flavors` requests. Also fixes tests for `list_servers` to pass again and add tests for `list_ssh_keys`.

### DIFF
--- a/lib/fog/digitalocean/requests/compute_v2/list_flavors.rb
+++ b/lib/fog/digitalocean/requests/compute_v2/list_flavors.rb
@@ -2,18 +2,18 @@ module Fog
   module Compute
     class DigitalOceanV2
       class Real
-        def list_flavors
+        def list_flavors(filters = {})
           request(
             :expects => [200],
             :method  => 'GET',
-            :path    => '/v2/sizes'
+            :path    => "/v2/sizes#{filters.to_a.map { |x| "#{x[0]}=#{x[1]}" }.join("&")}"
           )
         end
       end
 
     # noinspection RubyStringKeysInHashInspection
       class Mock
-        def list_flavors
+        def list_flavors(filters = {})
           response        = Excon::Response.new
           response.status = 200
           response.body   = {
--- a/lib/fog/digitalocean/models/compute_v2/flavors.rb
+++ b/lib/fog/digitalocean/models/compute_v2/flavors.rb
@@ -12,7 +12,7 @@ module Fog
         # @raise [Fog::Compute::DigitalOceanV2::ServiceError]
         # @see https://developers.digitalocean.com/documentation/v2/#list-all-sizes
         def all(filters = {})
-          data = service.list_flavors.body["sizes"]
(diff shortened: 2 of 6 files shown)


bsedat:
I'd love to consolidate the transformation for the filters into parameters somewhere. Not sure where that would be though. Happy to take suggestions :smile:

bsedat:
Should these changes be mirrored in the fog-digitalocean repo as well? What is the status of that? I'd love to switch over to it.

lib/fog/digitalocean/requests/compute_v2/list_flavors.rb
```diff
@@ -2,18 +2,18 @@ module Fog
   module Compute
     class DigitalOceanV2
       class Real
-        def list_flavors
+        def list_flavors(filters = {})
           request(
             :expects => [200],
             :method  => 'GET',
-            :path    => '/v2/sizes'
+            :path    => "/v2/sizes#{filters.to_a.map { |x| "#{x[0]}=#{x[1]}" }.join("&")}"
```
geemus: We may need/want to do some escaping of these key/values instead of just interpolating. What do you think?

geemus:
Hey, I can imagine consolidation being good there. Maybe a method on the main compute class that you could call out to in order to get the formatting or whatever.

I think fog-digitalocean was meant to replace this, but that it never quite managed to finish being extracted. Would be great to get it there. @jjasghar could you clarify perhaps?

Thanks!

geemus:
@bsedat oh, also it looks like there are some test failures here. Could you maybe rebase against master? I realize it merges cleanly already, but I think newer master actually contains fixes that would allow tests to run (and hopefully pass) properly). Thanks!

lib/fog/digitalocean/requests/compute_v2/list_flavors.rb
bsedat: I agree, that's definitely something I'll do when I consolidate the calls. Is `CGI.escape` the preferred library in fog for URL-escaping input?

bsedat:
@geemus I think this branch is up-to-date with master already unfortunately. I was having some trouble getting the entire test suite to run locally, but I was running `FOG_MOCK=true bundle exec shindont tests/digitalocean` to try and isolate the tests for this part of the gem.

lib/fog/digitalocean/requests/compute_v2/list_flavors.rb
geemus: Yeah, I think `CGI.escape` should be good

geemus:
Fair enough. I might need to tighten the fog-core dependency again...

bsedat:
Looks like we can pass the query parameters to Excon and that should take care of the escaping without us string-interpolating manually. I've proposed that change in #3894.

geemus:
Sounds good!

PR #3894, opened 2016-07-07
Digital Ocean list_* requests pass query params to Excon

bsedat:
Use Excon’s query parameter handling and escaping rather
than doing it manually.


bsedat:
This also fixes some embarrassing errors in my previous string interpolations 😢

geemus:
Cool, looks like a nice refinement. Thanks!

Issue #3905, opened 2016-08-26
Cloudstack create_tags request will fail with asterisk (*) character


gavinkflam:
Cloudstack create_tags request will fail with asterisk (*) character

In general, any requests with a asterisk (*) character is malformed. The deep rooted reason is that at cloudstack core, CGI escape is used. As of HTTP specification, we should be using URI escape instead.

gavinkflam:
#3906

gavinkflam:
Fixed with pull request #3906.

PR #3906, opened 2016-08-26
Use URI escape instead of CGI escape as of HTTP specification

gavinkflam:
In general, any requests with a asterisk (*) character is malformed.

The deep rooted reason is that at cloudstack core, CGI escape is used. As of HTTP specification, we should be using URI escape instead.

### DIFF
--- a/lib/fog/cloudstack/core.rb
+++ b/lib/fog/cloudstack/core.rb
@@ -13,6 +13,8 @@ module Fog
     def self.escape(string)
       string = CGI::escape(string)
       string = string.gsub("+","%20")
+      # Escaped asterisk will cause malformed request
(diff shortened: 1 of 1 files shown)


gavinkflam:
#3905

geemus:
I'm a bit reticent to fix it this way as URI.encode is marked as obsolete (see https://github.com/ruby/ruby/blob/trunk/lib/uri/common.rb#L98). I don't love it, but maybe we should just add an additional gsub for `*`  so that we can use the non-obsolete `CGI.escape`. What do you think?

gavinkflam:
@geemus Sure Wesley, I have just updated with a new commit
Thank you for your timely response

geemus:
Great, thanks!

PR #3909, opened 2016-09-06
Udpated gemspec, included fog-azure-rm as a dependency

bilal-naeem-confiz:
Added fog-azure-rm version 0.0.5 to fog.gemspec


plribeiro3000:
We can't accept this yet. `fog-azure` needs to support ruby `1.9` or `fog` will break for a lot of users.

plribeiro3000:
Also. I've raise this question elsewhere but got nothing so i will do it here again.
Could you please clean [this code](https://github.com/fog/fog-azure-rm/blob/master/lib/fog/azurerm.rb#L21-L41) and use fog DSL instead?

bilal-naeem-confiz:
Ok I'll implement fog DSL in place of the said code.
Also, it appears that Ruby v1.9.3 can be supported by modifying the source code to some extent. Previous versions of Ruby (i.e. < 1.9.3) cannot be supported due to conflicts arising in the dependent gems.

plribeiro3000:
Hmmm. Wouldn't you release a new version with only `Storage` support so we would not have to drop `1.9` support?
Also, Could you bump it to a `MAJOR` release instead, so we can know for sure its going to be a `Storage` only series?

plribeiro3000:
As we discussed this is not going to be accepted because it will break `fog` for users that for some reason still need to use ruby `1.9` so the solution we are going to follow is to leave `fog-azure-rm` as a standalone gem only until `fog` drop support for ruby `1.9`. After that we might add it to `fog` as a dependency if thats still the way we handle all the providers.

plribeiro3000:
I believe we can bring `fog-azure-rm` now.

Could you open another PR, please?

Thanks!

geemus:
Thanks!

bilal-naeem-confiz:
@plribeiro3000 @geemus 
Sure I'll open a PR first thing tomorrow

PR #3911, opened 2016-09-07
Add new api endpoint to fog for bulk updates

tameem92:


geemus:
Great, could you add a test around this as well? Thanks!

tameem92:
Got it. Will add test case around it.

PR #3912, opened 2016-09-13
Hp cleanup

voxik:
HPCloud support was removed in 817f85ca7b28b76fbd98bb0740960dd4342ea47e but it seems there are still some remnants which should be removed (I hope I did not used too big hammer to remove entire tests/hp directory structure).


geemus:
Good catch, I think this should be safe and remove some cruft. Thanks!

PR #3917, opened 2016-09-27
F rage4 bulk update endpoint

tameem92:


grossjo:
+1 if build passes.

geemus:
Thanks!

PR #3919, opened 2016-09-28
Revert "Move digitalocean to gem"

JurgensSmit:


geemus:
Hey, I think we still want this commit to be true (there are other things we need to fix in order to fix the build). Sorry for any confusion, and certainly let me know if I'm missing something here. Thanks!

PR #3920, opened 2016-09-28
Revert "Move digitalocean to gem"

majicaster2:


plribeiro3000:
Is there something wrong with this?

We just got another PR #3919 doing the same thing.

Whats the reason behind this?

plribeiro3000:
I'm closing this for now to make sure this does not get merged unintentionally.

Please, comment back with your reason to do this so we can trace the issue and figure out the best way to fix it.

Thanks!

PR #3922, opened 2016-09-28
Fix digitalocean bin issue

h0lyalg0rithm:
Sorry for introducing the error in the first place.This should fix it. :sob:


plribeiro3000:
Thats ok. =)

Regarding the tests it would be good to maintain them at least for now. Those were added by @tokengeek as a way to make sure the extracted providers are working as they should.
I believe we fixed this by adding the binary in `fog-digitalocen`.

h0lyalg0rithm:
Got it will try to fix it in the fog-digitalocean repo

plribeiro3000:
Sure. I believe @Temikus fixed it already. The issue now rely on `json` not being able to install. =s

h0lyalg0rithm:
Oh.Didnt look at travis.Seems like its still failing on my machine.probably some setup issue.

PR #3925, opened 2016-09-30
[vcloud_director] Add Edge_Gateways to  data model

tlawrence:
Adding the ability to view vShield Edge Routers via the model.

Not sure if the correct data is mocked for tests at this stage but it doesn't break anything else as far as I can tell..

I will look at tests next week to complement.


geemus:
Looks reasonable so far, will let it wait for tests for now, but let me know if you have questions or anything I can help with.

PR #3935, opened 2016-11-17
Fix typo in contributors

mostafaz:


plribeiro3000:
Hey @mostafaz. thanks for this catch.

I'm just not sure this will be overwritten by the tool used to generate it. This list is automatic generated on top of the git history and since it is incorrect it might come back. Perhaps @geemus can be more sure about it?

geemus:
Yeah, unfortunately we generate this, so direct changes would be overwritten. The only real "fix" would be to modify the history of the repo (as one or more commits must have come in with that set on the metadata). Rewriting history, however, seems too dangerous for an issue like this. You can also see the impact of this in how many different ways my name appears, for instance, as I didn't realize that it was different on different computers over time. Anyway, unfortunately I don't think there is a good way to fix this without moving away from generation, and if we move away from generation I have strong doubts we'll have any luck keeping this updated appropriately.

PR #3937, opened 2016-12-16
Add #deploy to VcloudDirector::Vm

harmjanblok:
To force customization on deployment the vCloud Director API provides an `forceCustomization` attribute in the `deploy` operation. The method `post_deploy_vapp` which can handle this operation (and accepts this attribute) has been present for many years (lib/fog/vcloud_director/requests/compute/post_deploy_vapp.rb). However currently the method cannot be accessed from a `VcloudDirector::Vm` or a `VcloudDirector::Vapp`. Hence this proposal to add this method in those classes.

During development it raised the following `TaskError`:

    Fog::Compute::VcloudDirector::TaskError: status: error, error: {:majorErrorCode=>"400", :message=>"[ a36d1a98-288a-46ce-9e64-3143c55a2c49 ] Parameter forceCustomization is not supported for vApps.", :minorErrorCode=>"BAD_REQUEST"}

Should I provide some documentation and testing on the `deploy` method? Are there any other optimizations I can or should add?

### DIFF
--- a/lib/fog/vcloud_director/README.md
+++ b/lib/fog/vcloud_director/README.md
@@ -410,7 +410,7 @@ vdc = org.vdcs.first
 vapp = vdc.vapps.get_by_name("segundo")
 vm = vapp.vms.get_by_name("DEVWEB")
 customization = vm.customization
-customization.compute_name = "NEWNAME"
+customization.computer_name = "NEWNAME"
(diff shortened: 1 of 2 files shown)


geemus:
Looks like a good start. I think some documentation and tests would round it out nicely. Thanks!

Issue #3942, opened 2017-02-10
Separate leftover providers


Temikus:
Ok, let's get this ball rolling, here's the list of providers that I've generated from lib:
`find . -maxdepth 1 -type f | sed 's/\.\//- [] /g'`

- [] bare_metal_cloud.rb
- [] bluebox.rb
- [] clodo.rb
- [] cloudsigma.rb
- [] cloudstack.rb
- [] dnsimple.rb
- [] dnsmadeeasy.rb
- [] dreamhost.rb
- [] fogdocker.rb
- [] glesys.rb
- [] go_grid.rb
- [] ibm.rb
- [] internet_archive.rb
- [] joyent.rb
- [] linode.rb
- [] ninefold.rb
- [] opennebula.rb
- [] openvz.rb
- [] ovirt.rb
- [] rage4.rb
- [] vcloud.rb
- [] vcloud_director.rb
- [] version.rb
- [] zerigo.rb

For every provider we need to make a decision:
- Separate
or
- Remove

@plribeiro3000 / @geemus PTAL
Also - can you give an idea (very general is ok, but at least something) on how should separation should be done ideally? I'll format it so it looks straightforward.

plribeiro3000:
Sure. I did a very basic document on it in the past. You can find it here:
https://github.com/fog/fog/wiki/Create-New-Provider-from-Scratch

Is that what you are looking for?

Em sex, 10 de fev de 2017 02:49, Artem Yakimenko <notifications@github.com>
escreveu:

lanej:
@Temikus nice work

lanej:
Looks like `fog-dnsimple` just needs removal from this gem as https://github.com/fog/fog-dnsimple already exists.  @weppos can you confirm?

plribeiro3000:
Good catch @lanej !

weppos:
Yep, I confirm @lanej. The DNSimple provider has been moved to fog-dnsimple.

geemus:
Awesome. If you are taking on something, please edit the comment above (or add a comment here) that says which provider, who you are and what date (so we can judge staleness). That way we don't step on one another's toes. I'll update the listing above with ones I know about.

geemus:
I also manually removed `bin.rb` and `version.rb` from the listing (just noting as the find/sed would no longer match).

lanej:
I think ninefold can be dropped entirely.

[...]

PR #3943, opened 2017-02-14
extract DNSimple, require 'fog-dnsimple'

lanej:
**BREAKING CHANGE**

`fog-dnsimple` uses the `Dnsimple` constant (conventional) where fog previously used `DNSimple`.  I could alias the constant but that doesn't change the `Fog.providers[:dnsimple]` value.

References: https://github.com/fog/fog/issues/3942


geemus:
Thanks for the effort and heads up on the potential breakage, hopefully it won't be too impactful (and it's a pretty easy fix).

Issue #3933, opened 2016-10-24
Flaky/Failing tests in Travis


Temikus:
Porting from #3930

Our Travis builds are failing.

I see 1 test consistently failing:

```
Failure:
Fog::Compute::when ecloud is passed with no available credentials#test_0001_returns ArgumentError [/home/travis/build/fog/fog/spec/fog/compute_spec.rb:13]:
ArgumentError expected but nothing was raised.
```

And 4 flaky ones:

```
Fog::Storage::when aliyun is passed with no available credentials#test_0001_returns ArgumentError [/home/travis/build/fog/fog/spec/fog/storage_spec.rb:9]:
ArgumentError expected but nothing was raised.
```

```
Fog::Compute::when aliyun is passed with no available credentials#test_0001_returns ArgumentError [/home/travis/build/fog/fog/spec/fog/compute_spec.rb:13]:
[ArgumentError] exception expected, not
Class: <URI::InvalidURIError>
Message: <"bad URI(is not URI?): ">
---Backtrace---
/home/travis/.rvm/rubies/ruby-2.0.0-p598/lib/ruby/2.0.0/uri/common.rb:176:in `split'
/home/travis/.rvm/rubies/ruby-2.0.0-p598/lib/ruby/2.0.0/uri/common.rb:211:in `parse'
/home/travis/.rvm/rubies/ruby-2.0.0-p598/lib/ruby/2.0.0/uri/common.rb:747:in `parse'
/home/travis/.rvm/gems/ruby-2.0.0-p598/gems/fog-aliyun-0.1.0/lib/fog/aliyun/compute.rb:200:in `initialize'
/home/travis/.rvm/gems/ruby-2.0.0-p598/gems/fog-core-1.43.0/lib/fog/core/service.rb:108:in `new'
/home/travis/.rvm/gems/ruby-2.0.0-p598/gems/fog-core-1.43.0/lib/fog/core/service.rb:108:in `new'
/home/travis/.rvm/gems/ruby-2.0.0-p598/gems/fog-core-1.43.0/lib/fog/core/services_mixin.rb:16:in `new'
/home/travis/.rvm/gems/ruby-2.0.0-p598/gems/fog-core-1.43.0/lib/fog/compute.rb:54:in `new'
/home/travis/.rvm/gems/ruby-2.0.0-p598/gems/fog-core-1.43.0/lib/fog/core/services_mixin.rb:4:in `[]'
[...]

PR #3945, opened 2017-02-23
Fix dependency issues when building

plribeiro3000:
Ok. Now all builds succeed to install all dependencies but we still have some broken tests but we can use #3933 for that. =)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,12 +5,9 @@ sudo: false
 script: bundle exec rake travis
 
 rvm:
-  - 1.9
-  - 2.0
   - 2.1
   - 2.2
   - jruby-head
-  - jruby-19mode
 
 gemfile:
   - Gemfile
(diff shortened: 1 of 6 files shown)


plribeiro3000:
I'm gonna go ahead and merge since this is just a build change, mostly on travis.

plribeiro3000:
cc/ @Temikus @icco @geemus @tokengeek Do you know where does the dependency `public_suffix` come?

I did install old ruby versions on my machine and did a clean install without it but travis keep complaining it could not install it so i had to force its version. =s

geemus:
afraid I'm not sure off hand, gemfile lock should maybe show in some way?

plribeiro3000:
I thought the same. The issue is that locally it does not install it but in
Travis it does. Is there a way to see the build in Travis (I mean the
artifacts and files)?

Em sex, 24 de fev de 2017 12:17, Wesley Beary <notifications@github.com>
escreveu:

icco:
@plribeiro3000 you could always add `cat Gemfile.lock` to the script block. You could also get complicated with https://docs.travis-ci.com/user/uploading-artifacts/

PR #3948, opened 2017-03-08
Remove internet_archive

icco:
For #3942, dependent on https://github.com/fog/fog-internet-archive being working.


icco:
This should be ready to go @geemus

icco:
Or maybe not...

geemus:
@icco looks like a couple more lingering weird things around bin still left, if you could fix those also that would be awesome. Thanks!

icco:
Ok @geemus, looks fixed.

PR #3954, opened 2017-03-31
[ovirt] wait for stopped before starting when rebooting

iNecas:
When stopping and starting as part of the reboot, it can happen the VM
stopping is not finished before starting, leading to getting this error
in oVirt:

[Operation Failed: [Cannot run VM because the VM is in Powering Down status.]

### DIFF
--- a/lib/fog/ovirt/models/compute/server.rb
+++ b/lib/fog/ovirt/models/compute/server.rb
@@ -44,7 +44,7 @@ module Fog
         end
 
         def stopped?
-          !!(status =~ /down/i)
+          status.downcase == 'down'
         end
 
         def mac
@@ -134,7 +134,10 @@ module Fog
         end
 
         def reboot(options = {})
-          stop unless stopped?
+          unless stopped?
+            stop
+            wait_for { stopped? }
+          end
           start options.merge(:blocking => true)
         end
 


geemus:
@iNecas Will this still work ok with new servers? ie do they also respond in a truthy way to stopped? Alternatively, are these just not methods you would use with new servers? Just want to try and ensure we don't break one thing when trying to fix another. Thanks!

bkearney:
@iNecas bump on this.

iNecas:
@geemus after more testing, I decided to put it directly into the reboot method, as it's mainly applicable there. Also, the more sever thing is the `stopped?` method was not working properly with RHEV 4, as they have `powering_down` state there, which is of making the `stopped?` method return true sooner, than it actually should

lib/fog/ovirt/models/compute/server.rb
```diff
@@ -44,7 +44,7 @@ def locked?
         end
 
         def stopped?
-          !!(status =~ /down/i)
+          !!(status =~ /^down$/i)
```
tbrisker: not significant, but doesn't `status.downcase == 'down'` have the same effect while being faster and easier to read?

iNecas: Good point, updated

iNecas:
test failures unrelated

orrabin:
This fix in `stopped?` also fixes another bug in `destroy` not waiting for the vm to stop (http://projects.theforeman.org/issues/15441).

geemus:
Sorry for the delay (was on vacation last week). LGTM. Thanks!

iNecas:
Thanks @geemus, any chance to get a new fog version released, as there are several bugfixes in ovirt provider that we would like to get to our users?

geemus:
Sorry for (further) delay, should now be available as v1.41.0. Thanks!

PR #3956, opened 2017-04-07
Joyent Extraction

effeminate-batman:
This is the Joyent extraction PR to remove the Joyent provider from Fog.


icco:
cc @geemus and #3942

geemus:
Awesome.

@effeminate-batman could you add me as an owner to the fog-joyent gem? That way there is some fallback and we can do security or other emergency patches if needed.

geemus:
Thanks!

PR #3957, opened 2017-05-07
Update fog.gemspec

ryuheechul:
use recent fog-goole to support GCS JSON api


icco:
Sorry, we can't support this because fog-google only supports Ruby 2.0 and fog still supports 1.8. please just require fog-google in your project.

PR #3958, opened 2017-05-14
[ovirt] adding disks and clone to attribute

orrabin:
This is to be able to pass `disks` and `clone` to become a part of the xml created here: https://github.com/abenari/rbovirt/blob/master/lib/ovirt/vm.rb#L111-L117

See comment: https://github.com/theforeman/foreman/pull/4512#issuecomment-301305759

### DIFF
--- a/lib/fog/ovirt/models/compute/server.rb
+++ b/lib/fog/ovirt/models/compute/server.rb
@@ -31,6 +31,8 @@ module Fog
         attribute :ips
         attribute :ha
         attribute :ha_priority
+        attribute :clone
+        attribute :disks
 
         def ready?
           !(status =~ /down/i)


icco:
@orrabin seems reasonable. Did you test this on a working ovirt deployment anywhere?

icco:
Oh also, does this require a certain version of `rbovirt`?

orrabin:
I tested this with `ovirt` 3.6, `rbovirt` 0.1.3 but I don't think the versions really matter on this, anything in the `attributes` will become a symbol in the hash of options sent to the xml creation in rbovirt which will create the expected xml attribute.
The current situation is when I try to add `clone` and `disks` they are ignored and don't appear in the xml.

lib/fog/ovirt/models/compute/server.rb
```diff
@@ -31,6 +31,8 @@ class Server < Fog::Compute::Server
         attribute :ips
         attribute :ha
         attribute :ha_priority
+        attribute :clone
```
iNecas: Alternatively, we could define clone_attributes as methods, that would do the mapping between how one would want to define the clone attributes, and how rbovirt maps it (as the way the rbovirt does it is not very nice).

```
def clone_attributes
  extract_clone_attributes(@attributes)
end

# Accepted keys: :clone, :disks
def clone_attributes=(clone_attributes)
  @attributes.merge!(extract_clone_attributes(clone_attributes)
end

def extract_clone_attributes(attributes)
  ret = {}
  ret[:clone] = attributes[:clone] if attributes.key?(:clone)
  ret[:disks] = attributes[:disks] if attributes.key?(:disks)
  ret
end
```

And in the vm options, we would pass `:clone_attributes => { :clone => true, :disks => [...]}`

geemus: Good point. It seems like we could add this on top of the current patch pretty easily without issue, so I'm going to go ahead and accept this and we can add that later if desired.

geemus:
Thanks!

PR #3960, opened 2017-07-07
[ovirt] Attach ovirt activate and deactivate volume requests

karcaw:
These two calls are available in the rbovirt library, and I needed access to them.


geemus:
Thanks!

PR #3971, opened 2017-09-28
Upgrade JSON to ~> 2.0

ddfreyne:
The README states

> Fog requires Ruby 2.0.0 or later.

and `json` 2.0 should thus now work properly.

(Also see #3707, which removed Ruby 1.9.3 support.)


geemus:
Thanks!

PR #3974, opened 2017-10-27
Drop Ruby<2 support

tbrisker:
Once this is merged, version 2.0 should be cut and the 1.x branch should only receive critical security fixes.


geemus:
Thanks!

PR #3980, opened 2018-02-07
add id in xml only if present

tinguelyp:
add id in xml only if present. Else an id is generated from xsd generation


geemus:
Thanks!

PR #3981, opened 2018-02-16
make network optionnal for instantiate request

tinguelyp:
- if we don't want to pass a network while the instantiation we have to make this field optionnal
- implement the put api to configure the network section in vapp


geemus:
Thanks!

PR #3983, opened 2018-02-25
Drop Zerigo

plribeiro3000:
Zerigo is dead. https://news.ycombinator.com/item?id=13863592


plribeiro3000:
Merging since the broken test is unrelated to this change.

geemus:
Thanks!

PR #3984, opened 2018-03-02
Introduce kubevirt provider

pkliczewski:
We update provider dependencies to enable [kubevirt](http://www.kubevirt.io/) provider.


geemus:
Looks like fog-kubevirt hasn't been released yet? Also, if you add me as an administrator on your repo, I can move it into the fog org and get that part sorted out for you. I'd like to see both of those things done before we merge further things. Thanks!

pkliczewski:
@geemus I transferred the repository to you, I am not able to give admin to my personal org.

geemus:
@pkliczewski ah cool, yeah I probably gave slightly incorrect instructions (it's been a while since I've initiated a transfer). I think that should work though, will let you know if I need anything else. Thanks!

geemus:
@pkliczewski I think it should be transferred and you are added as admin. I only added you and @plribeiro3000 but I think as admin you can add others if you need or want to. Just let me know if you need anything else on that front. Thanks!

lib/fog/kubevirt.rb
```diff
@@ -0,0 +1 @@
+require 'fog/kubevirt/compute'
```
masayag: this should be modified to `fog/compute/kubevirt'` to match https://github.com/fog/fog-kubevirt/blob/master/lib/fog/compute/kubevirt.rb

pkliczewski:
We still want to continue to work on it. Can we have it reopen or should I create new PR?

geemus:
Sorry, I had thought we resolved this from the discussion, but my memory is faulty at times. Reopening is fine. Thanks!

PR #3989, opened 2018-04-19
[cloudstack] Allow specifying the size of root volume

k0kubun:
Similar to https://github.com/fog/fog/pull/3566

Not only the size of DATADISK, but also I would like to specify the size of root disk for cloudstack.

http://cloudstack.apache.org/docs/api/apidocs-4.4/root_admin/deployVirtualMachine.html


plribeiro3000:
Awesome. Thanks!

PR #3991, opened 2018-04-19
Ruby 2.5 compatibility

pvalena:
[CGI::escape](https://github.com/fog/fog/blob/master/lib/fog/cloudstack/core.rb#L14) in Ruby 2.5 behaves differently.

I do not know how to make it pass for Ruby < 2.5 at the same time.

```
irb(main):011:0> RUBY_VERSION
=> "2.5.1"
irb(main):012:0> CGI::escape("'Stöp!' said Fred_-~.")
=> "%27St%C3%B6p%21%27+said+Fred_-~."
```

### DIFF
--- a/tests/cloudstack/signed_params_tests.rb
+++ b/tests/cloudstack/signed_params_tests.rb
@@ -1,7 +1,7 @@
 # encoding: utf-8
 
 Shindo.tests('Cloudstack | escape', ['cloudstack']) do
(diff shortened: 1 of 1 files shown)


plribeiro3000:
We need to check the ruby version at runtime and behave differently based on that.

Would you be able to tackle that?

pvalena:
Yes, sure.

PR #3997, opened 2018-06-19
Fix main Travis issues

tokengeek:
Hi guys,

I've dug into the errors I'm seeing on Travis and seem to have isolated
workarounds for the bulk of them.

This is still blocked by https://github.com/fog/fog-aliyun/pull/20 however
but there is a fix for that as well.

Individual commits should have details for each change that was needed
and the reasoning.

Ideally for the `opennebula` changes, there would be a `fog-core` error covering
"missing/optional by provider" dependencies rather than just `LoadError` so it
is clearer what the issue is.

It was only after digging into the "OpenNebula does not have a compute service"
problem did I realise `fog-core`'s `LoadError` handling was hiding a reference to
`xmlrpc` which was leaking from any call to `Fog::Compute[:opennebula]`

However, I didn't want to add an update to fog-core in here as well.

### DIFF
--- a/lib/fog/opennebula/compute.rb
+++ b/lib/fog/opennebula/compute.rb
@@ -98,7 +98,12 @@ module Fog
           @opennebula_endpoint = options[:opennebula_endpoint]
           @opennebula_username = options[:opennebula_username]
           @opennebula_password = options[:opennebula_password]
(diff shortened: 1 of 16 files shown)


geemus:
Thanks for digging in to this!

PR #3998, opened 2018-06-21
Fix deprecation warning appearing in tests

tokengeek:
This fixes a number of references to deprecated methods or classes.

The main bulk is removing the tests for the Fog "binary" that allowed providers to call their services directly from the top level, e.g. `Brightbox[:compute]`

These tests would flood the testing output with deprecation warnings when no one will hopefully be using that interface anymore.

It also fixes a reference to `#connection` which should be `#service`.

The warnings have disappeared on Travis but there are still plenty more.

### DIFF
--- a/spec/fog/bin/atmos_spec.rb
+++ b/spec/fog/bin/atmos_spec.rb
@@ -20,14 +20,4 @@ describe Atmos do
       end
     end
   end
-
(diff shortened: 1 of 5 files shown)


geemus:
@tokengeek sounds good. And welcome back! (seems like it's been a while)

PR #4001, opened 2018-06-23
Add SemVer stability badge to README

greysteil:
First of all, thanks for fog!

Would you be up for adding a badge that shows how SemVer compliant / bug free new releases are? I was looking through the data we gather at Dependabot and realised we could put one together, so threw together the below:

[SemVer](https://dependabot.com/compatibility-score.html?dependency-name=fog&package-manager=bundler&version-scheme=semver)

If you click through then there's a description of how it's calculated - basically it takes all of the relevant updates Dependabot has done for projects that use fog and checks what percentage of the time specs pass on the upgrade PR.

The score is slightly lower than 100% because some projects have flaky specs, but I'm working on filtering them out from the data, too :octocat:.

Finally, I've removed the old Gemnasium badge, as it now just serves an advert for GitLab.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,9 +7,9 @@ fog is the Ruby cloud services library, top to bottom:
 * Mocks make testing and integrating a breeze.
 
 [![Build Status](https://secure.travis-ci.org/fog/fog.svg?branch=master)](http://travis-ci.org/fog/fog)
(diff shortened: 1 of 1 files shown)


plribeiro3000:
Wow. Thats a cool feature to have.

@geemus Any ideas/thoughts?

geemus:
Sure, happy to give it a try. Thanks!

PR #4002, opened 2018-06-26
Removing leftover file from fog-ovirt extraction

orrabin:
@plribeiro3000 This is what we talked about regarding:https://github.com/fog/fog-ovirt/pull/29#issuecomment-399957348.
I will open another PR against master that also removed 2 more leftover files I found (and open a PR to `fog-ovirt` to add them).

PR #4003, opened 2018-06-26
Removing leftovers from extraction of fog-ovirt

orrabin:
Looks like activate/deactivate volume already exist in `fog-ovirt` but they were not deleted from fog.
Also removing `lib/fog/ovirt.rb` as in: https://github.com/fog/fog/pull/4002


geemus:
Thanks!

PR #4004, opened 2018-07-07
Packet provider support

jasmingacic:
Packet provider support


geemus:
Glad to see the progress here. That said, is there a particular need to add this to the dependency list for fog/fog? We've started trying to move away from that pattern, due to dependency conflicts and related issues that tend to creep in over time. I think the better long term solution is likely for people to just include the providers that they want, but it's been a slow process for us to extract the stuff that is already in there. Does that make sense? What do you think? Thanks again for your work!

PR #4010, opened 2018-09-14
Remove BlueBox Blocks

ChrisLundquist:
https://www.geekwire.com/2015/ibm-acquires-seattles-blue-box-in-exit-for-private-cloud-startup/
BlueBox was acquired by IBM and they moved towards OpenStack.
While I no longer work at BlueBox an IBM Company, I wrote large parts of
the server side and to the best of my knowledge this offering is no
longer available.

To the former BlueBoxers:
.cc @jesseproudman @leehuffman @TrevorBramble @scoop206 @dlundquist


scoop206:
👍 dust in the wind

Thanks @ChrisLundquist

jesseproudman:
End of an era. Thanks gents.

PR #4013, opened 2018-09-24
lock xenserver version

### LINKED ISSUE #4012: Unpinned version of fog-xenserver breaks with older ruby versions
Greetings, 

It looks like a recent release of `fog-xenserver` has a new dep on `xmlrpc` which requires Ruby version >= 2.3. 

Would it be possible to add some version constraints?
https://github.com/fog/fog/blob/v1.37.0/fog.gemspec#L73

```
# /opt/chef/embedded/bin/gem install fog --no-rdoc --no-ri -v "1.37.0"
ERROR:  Error installing fog:
    xmlrpc requires Ruby version >= 2.3.
```

ruby: 2.1.9p490 (2016-03-30 revision 54437) [x86_64-linux]

https://github.com/fog/fog-xenserver/issues/74

geemus:
fixes #4012

### DIFF
--- a/fog.gemspec
+++ b/fog.gemspec
@@ -82,7 +82,7 @@ Gem::Specification.new do |s|
   s.add_dependency("fog-vmfusion")
   s.add_dependency("fog-voxel")
   s.add_dependency("fog-vsphere", ">= 0.4.0")
-  s.add_dependency("fog-xenserver")
+  s.add_dependency("fog-xenserver", "<= 0.3.0")
 
   s.add_development_dependency("docker-api", ">= 1.13.6")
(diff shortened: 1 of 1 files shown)


geemus:
@powellchristoph - I think this is what you were asking for, can you confirm? Thanks!

plribeiro3000:
Hmmmm. Can't you use a recent version of `fog`? Like `1.42.1`?

This one is to old and the problem you describe should be fixed at the latest.

@geemus I think that should fix it but i wonder if this wont be fixed with bundle update as well.

plribeiro3000:
Im closing this due to lack of response.

PR #4015, opened 2018-10-16
lib/fog/linode: Add Tokyo2 to avail_datacenters

scub:
It would appear that the existing test cases should suffice for providing coverage: https://github.com/fog/fog/blob/master/tests/linode/requests/compute/datacenter_tests.rb


scub:
Hey friends,

Looks like we've got a few test failures however they appear to be in present builds and unrelated to this particular change. Any advice on how to proceed?

Cheers!

geemus:
@scub thanks for the fix and questions. We should fix the tests, but as you suggest, your change shouldn't relate in any way. I'll happily bring it in. Thanks!

PR #4018, opened 2018-11-07
Fix compatibility with fog-brightbox 1.0.0+.

voxik:
fog-brightbox updated namespaces into Fog::Brightbox format.

I am not sure if this is compatible with older fog-brightbox, which is now demanded by fog, but I guess I'll know soon :)

### DIFF
--- a/lib/fog/bin/brightbox.rb
+++ b/lib/fog/bin/brightbox.rb
@@ -3,9 +3,9 @@ class Brightbox < Fog::Bin
     def class_for(key)
       case key
       when :compute
-        Fog::Compute::Brightbox
+        Fog::Brightbox::Compute
       when :storage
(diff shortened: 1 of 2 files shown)


voxik:
Related to this commit https://github.com/fog/fog-brightbox/commit/bf49d73d9ad866b19952cf9d8f4ac9ba8db0a854

voxik:
I don't see any brightbox related issues in the travis output, so hopefully the patch is compatible with older versions of the package.

geemus:
Looks good to me.

@tokengeek could you confirm? Thanks!

tokengeek:
Yep. It's difficult to confirm with the other errors but nothing serious seems to be turning up using either of the two versions of `fog-brightbox` (0.16.1 and 1.0.0)

PR #4020, opened 2019-06-13
Removed all references to fog-cloudstack.

kovukono:
Fog-cloudstack can be added back into lib/fog/bin.rb once we have a fog-cloudstack gem.


geemus:
Looks good, thanks!

geemus:
Thanks!

PR #4026, opened 2020-03-23
json-dos-cve-2020-10663

onecareer-yamamoto:
https://www.ruby-lang.org/en/news/2020/03/19/json-dos-cve-2020-10663/

Upgreade json gem version


onecareer-yamamoto:
@geemus 

>  Could you please omit the version bump for the fog gem itself? I'll do that when I release

Sorry for not considerate release operation.

Remove version bump and rebased 8e1d85d

geemus:
Thanks for the fix and also for accommodating my request.

PR #4038, opened 2021-06-03
Supported Ruby version is 2.5+ now

lzap:
Ruby 2.0 is EOL, supported Ruby versions are 2.6+ and currentl Rubocop supports 2.4+. This is a good compromise.


geemus:
Thanks! The test fail is unrelated (and looks to be an issue with a community lib I haven't touched previously). I'll work on a distinct fix for that presently.

geemus (approved):

PR #4069, opened 2022-04-06
Set permissions for GitHub actions

naveensrinivasan:
- Included permissions for the action. https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions

https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

[Keeping your GitHub Actions and workflows secure Part 1: Preventing pwn requests](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)

 Restrict the GitHub token permissions only to the required ones; this way, even if the attackers will succeed in compromising your workflow, they won’t be able to do much.

Signed-off-by: naveensrinivasan <172697+naveensrinivasan@users.noreply.github.com>

### DIFF
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -13,6 +13,11 @@ on:
   pull_request:
     branches: [ master ]
 
+permissions:
+  contents: read
+
 jobs:
   test:
+    permissions:
+      contents: none
     uses: fog/.github/.github/workflows/ruby.yml@v1.1.0
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -4,6 +4,11 @@ on:
   schedule:
   - cron: "30 1 * * *"
 
+permissions:
+  contents: read
+
 jobs:
   stale:
+    permissions:
+      contents: none
     uses: fog/.github/.github/workflows/stale.yml@v1.1.0


geemus:
@naveensrinivasan Thanks.

That being said, I attempted at least to do this at the shared workflow level. I'd love your eyes to double check I did it correctly: https://github.com/fog/.github/commit/f9db19967a5a11c1f962ee1b81d6c2af619602ec

I that is correct, then I think we can just bump the tags on the shared things here to v1.2.0 and be done? I thought that might simplify and make it easier to roll out that change consistently across the many fog related repos.

Does that make sense? I certainly welcome any feedback and questions.

naveensrinivasan:
> @naveensrinivasan Thanks.
> 
> That being said, I attempted at least to do this at the shared workflow level. I'd love your eyes to double check I did it correctly: [fog/.github@f9db199](https://github.com/fog/.github/commit/f9db19967a5a11c1f962ee1b81d6c2af619602ec)
> 
> I that is correct, then I think we can just bump the tags on the shared things here to v1.2.0 and be done? I thought that might simplify and make it easier to roll out that change consistently across the many fog related repos.
> 
> Does that make sense? I certainly welcome any feedback and questions.

Cool, it looks good. 👍

naveensrinivasan:
> > @naveensrinivasan Thanks.
> > That being said, I attempted at least to do this at the shared workflow level. I'd love your eyes to double check I did it correctly: [fog/.github@f9db199](https://github.com/fog/.github/commit/f9db19967a5a11c1f962ee1b81d6c2af619602ec)
> > I that is correct, then I think we can just bump the tags on the shared things here to v1.2.0 and be done? I thought that might simplify and make it easier to roll out that change consistently across the many fog related repos.
> > Does that make sense? I certainly welcome any feedback and questions.
> 
> Cool, it looks good. 👍

I didn't find documentation if the shared workflows reuse the permissions. The permissions defined in the shared workflow flows down to this one. Just to be safe I would add permissions in them. As it is not an on-going maintenance it shouldn't be that bad. That's my two cents. I'll let you decide.

geemus:
@naveensrinivasan - So to clarify, you suggest we would add the permissions to the shared workflow, but also here as well? I think having it as the default for the workflows makes sense, but it could leave other things vulnerable (because someone could change what that tag points at and change all the permissions at once). So the safer (and more clear/explicit) additional step would be to put it here as well. Does that line up with what you were thinking?

naveensrinivasan:
> So the safer (and more clear/explicit) additional step would be to put it here as well. Does that line up with what you were thinking?

Yes!

geemus:
@naveensrinivasan Got it. Sounds good, thanks for talking through it with me.

PR #4081, opened 2022-10-25
fix aws

manutazo:


geemus (changes requested):
Hey, thanks for the fix to the naming order. I think the service_mapper and aws_spec changes look good, but I would prefer to leave the changelog and gemspec as they are. Could you remove them from the PR please?

manutazo:
I'm sorry, it was an oversight on my part, is it correct now?

manutazo:
I'm sorry, it was an oversight on my part, is it correct now?

geemus (approved):

PR #4082, opened 2022-10-25
add libcurldev to github actions setup for fog-aliyun

geemus:


geemus (commented):
Doesn't wholly fix tests, but fixes some so it's progress.

PR #4095, opened 2023-08-10
CI against Ruby 3.2

y-yagi:
It seems CI isn't stable now.
But this setting affects to CI of `fog-core`. So I would like to add this for that.
https://github.com/fog/fog-core/blob/75b972e0d241a4e40911c1003643605533dba26b/.github/workflows/ruby.yml#L18

### DIFF
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -26,7 +26,7 @@ jobs:
 
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


geemus:
@y-yagi Sounds good. Yes, unfortunately due to the large network of dependencies it's quite difficult to keep this shared set of tests working.

